PR #14, opened 2017-01-18
Adds a convenience parameter to FieldOperator

andyHa:
Adds a convenience parameter to FieldOperator

Many queries against magnitudes or dates look like
WHERE (field > $param OR field IS NULL)

 This can now be achived using:
 FieldOperator.on(field).greateThan(param).orNull()


sabieber:
Please increase lib version

PR #19, opened 2017-02-16
When we JOIN-FETCH a field which has an alias (the column name in the database does not match the field name) we have to change the aliased column name accordingly.

andyHa:
When we JOIN-FETCH a field which has an alias (the column name in the database does not match the field name) we have to change the aliased column name accordingly.

If we join fetch a field which is named ‘address_name1’, with a column name ‘name1’ for table t1,
we have to generate SELECT … name1 AS t1_name1 instead of name1 AS t1_address_name1,
as the readEntity call will look for that column, when executing join fetches

### DIFF
--- a/src/main/java/sirius/db/mixing/SmartQuery.java
+++ b/src/main/java/sirius/db/mixing/SmartQuery.java
@@ -579,21 +579,29 @@ public class SmartQuery<E extends Entity> extends BaseQuery<E> {
             Tuple<String, EntityDescriptor> joinInfo = c.determineAlias(field.getParent());
             c.getSELECTBuilder().append(joinInfo.getFirst());
             c.getSELECTBuilder().append(".");
-            if (Entity.ID.getName().equals(field.getName()) || Entity.VERSION.getName().equals(field.getName())) {
-                c.getSELECTBuilder().append(field.getName());
-            } else {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mixing/SmartQuery.java
```diff
@@ -579,23 +579,31 @@ private void appendFieldList(Compiler c, boolean applyAliases) {
...
-                    c.getSELECTBuilder().append(field.getName());
+                    c.getSELECTBuilder().append(columnName);
                 }
                 c.createJoinFetch(field);
             }
         }
     }
 
+    private String fetchEffectiveColumnName(Column field, Tuple<String, EntityDescriptor> joinInfo) {
+        String columnName = null;
```
meg1502: Remove this useless assignment to local variable "columnName"


> A dead store happens when a local variable is assigned a value, including null, that is not read by any subsequent instruction. Calculating or retrieving a value only to then overwrite it or throw it away, could indicate a serious error in the code. Even if it's not an error, it is at best a waste of resources.
Even assigning null to a variable is a dead store if the variable is not subsequently used. Assigning null as a hint to the garbage collector used to be common practice, but is no longer needed and such code should be eliminated.

PR #39, opened 2017-04-11
SE-2002 Adds missing NLS Key

PhilHup:
Adds NLS keys for "EntityRefProperty.cannotDeleteEntityWithChildren" in the german and english properties.
Bumps version Number.

- fixes: SE-2002


src/main/resources/db_en.properties
```diff
@@ -24,6 +24,8 @@ Property.illegalValue=Invalid value '${value}' in field '${property}'.
 Property.fieldNotNullable=THe field '${field}' must not be empty.
 Property.fieldNotUnique=The value '${value}' is already in use for the field '${field}'.
 
+EntityRefProperty.cannotDeleteEntityWithChildren=The Entity referenced in the field '${field}' can not be deleted. There are '${count}' child Entities of the type '${type}' who reject this change.
```
andyHa: which reject this change

PR #40, opened 2017-04-12
Method for easy check if column was changed

fkrone:
It is now easier to check if the value of a column was changed after save. This new method provides easier access to this information.

### DIFF
--- a/src/main/java/sirius/db/mixing/Entity.java
+++ b/src/main/java/sirius/db/mixing/Entity.java
@@ -255,13 +255,18 @@ public abstract class Entity extends Mixable {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mixing/Entity.java
```diff
@@ -254,6 +254,16 @@ public String getIdAsString() {
         return String.valueOf(getId());
     }
 
+    /**
+     * Determines if the given {@link Column} was changed in this {@link Entity} since it was last fetched from the database.
+     *
+     * @param columnToCheck the column to check if if was changed
+     * @return <tt>true</tt> if the column was changed, <tt>false</tt> otherwise
+     */
+    public boolean isColumnChanged(Column columnToCheck) {
```
andyHa: Column... und ver-odern

PR #43, opened 2017-04-24
Repairs the generated query for MySQL

andyHa:
Although it should not matter, what is put into the select MySQL
behaves differently. So we use NULL, which works and clearly signals,
that we only want the existence to be checked and are not interested
in the fields.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.7.1</version>
     </parent>
     <artifactId>sirius-db</artifactId>
-    <version>3.0</version>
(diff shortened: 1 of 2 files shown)


andyHa:
This seems to be a MySQL bug...

Traditionally, an EXISTS subquery starts with SELECT *, but it could begin with SELECT 5 or SELECT column1 or anything at all. MySQL ignores the SELECT list in such a subquery, so it makes no difference.
https://dev.mysql.com/doc/refman/5.7/en/exists-and-not-exists-subqueries.html

PR #44, opened 2017-05-02
Distinguishes between null and 0 (entity ids)

tbiScireum:
By checking if a fetched id is null before assigning it we are able to differ between new objects (not found) and found ones with the id 0.
Also adds a before save check to the doUpdate method which checks if an entity ref points to a not saved entity.

### DIFF
--- a/src/main/java/sirius/db/mixing/EntityDescriptor.java
+++ b/src/main/java/sirius/db/mixing/EntityDescriptor.java
@@ -826,7 +826,9 @@ public class EntityDescriptor {
     private void readIdAndVersion(String alias, Set<String> columns, ResultSet rs, Entity entity) throws SQLException {
         String idColumnLabel = getIdColumnLabel(alias);
         if (columns.contains(idColumnLabel.toUpperCase())) {
-            entity.setId(rs.getLong(idColumnLabel));
+            if (rs.getString(idColumnLabel) != null) {
+                entity.setId(rs.getLong(idColumnLabel));
+            }
         }
         if (isVersioned()) {
             String versionColumnLabel = getVersionColumnLabel(alias);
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/db/mixing/OMA.java
```diff
@@ -148,6 +148,13 @@ public boolean isReady() {
             EntityDescriptor ed = entity.getDescriptor();
             ed.beforeSave(entity);
 
+            for (Property property : ed.getProperties()) {
```
andyHa: Nope, just fix EntityRef#containsNonpersistenValue / setValue to handle "new" values properly

src/main/java/sirius/db/mixing/EntityDescriptor.java
```diff
@@ -826,7 +826,9 @@ public Entity readFrom(String alias, Set<String> columns, ResultSet rs) throws E
     private void readIdAndVersion(String alias, Set<String> columns, ResultSet rs, Entity entity) throws SQLException {
         String idColumnLabel = getIdColumnLabel(alias);
         if (columns.contains(idColumnLabel.toUpperCase())) {
-            entity.setId(rs.getLong(idColumnLabel));
+            if (rs.getString(idColumnLabel) != null) {
+                entity.setId(rs.getLong(idColumnLabel));
```
andyHa: nope, use rs.wasNull() - *horror API* 

http://stackoverflow.com/questions/2920364/checking-for-a-null-int-value-from-a-java-resultset

src/test/java/sirius/db/mixing/SmartQuerySpec.groovy
```diff
@@ -25,6 +26,9 @@ class SmartQuerySpec extends BaseSpecification {
     @Part
     static OMA oma
 
+    SmartQueryTestParentEntity parent1
```
andyHa: do we need global variables here?

PR #47, opened 2017-05-30
Unique index names for derby

andyHa:


src/main/java/sirius/db/mixing/schema/DatabaseDialect.java
```diff
@@ -179,4 +179,13 @@
      * @return <tt>true</tt> if the key should be dropped, <tt>false</tt> otherwise
      */
     boolean shouldDropKey(Table targetTable, Table currentTable, Key key);
+
+    /**
+     * Computes the effective name used fora key.
```
sabieber: fora -> for a

PR #56, opened 2017-08-24
Adds a public method to check if Redis is still running.

D3nnisH:
Applications can call it prior to the actual Redis call to prevent `HandledExceptions` caused by `JedisConnectionExceptions` to be thrown at this place:

```
if (redis.isRunning()) {
    return redis.query(...);
} else {
    return null;
}
```


D3nnisH:
keep open pls

PR #64, opened 2017-10-24
Always fetch join columns

Tobsucht:
Always fetch the Id of the join columns as they are placed in the EntityRef implementation and a call to isFilled() may signal a wrong state, if a query was build using .fields() and the join-column was omitted.

### DIFF
--- a/src/main/java/sirius/db/mixing/SmartQuery.java
+++ b/src/main/java/sirius/db/mixing/SmartQuery.java
@@ -33,6 +33,8 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/db/mixing/SmartQuery.java
```diff
@@ -170,11 +172,86 @@ public EntityDescriptor getEntityDescriptor() {
...
+
+        return presentColumnPaths;
+    }
+
+    /**
+     * Synchonizes the columns which where explicitly set using {@link #fields(Column...)} with the columns that are
+     * required for join-fetches. If columns are missing, they are automatically added.
+     */
+    private void checkJoinColumnsPresent() {
+        Set<String> presentColumnPaths = aggregatePresentColumns();
```
andyHa: couldn't we build a single stream to do all this using map instead of severyl intermediate data structures?

src/main/java/sirius/db/mixing/SmartQuery.java
```diff
@@ -170,11 +172,86 @@ public EntityDescriptor getEntityDescriptor() {
...
+     */
+    private Set<String> aggregatePresentColumns() {
+        Set<String> presentColumnPaths = new HashSet<>();
+
+        this.fields.forEach(field -> {
+            Monoflop mf = Monoflop.create();
+            StringBuilder fieldName = new StringBuilder();
+            Column currentField = field;
+
+            while (currentField != null) {
```
andyHa: column.toString does all this(?)

src/main/java/sirius/db/mixing/SmartQuery.java
```diff
@@ -170,11 +172,86 @@ public EntityDescriptor getEntityDescriptor() {
...
+     */
+    private HashMap<String, Column> aggregateRequiredJoinColumns() {
+        HashMap<String, Column> requiredJoinFields = new HashMap<>();
+
+        this.fields.forEach(field -> {
+            Monoflop mf = Monoflop.create();
+            StringBuilder fieldName = new StringBuilder();
+            Column currentField = field.getParent();
+
+            while (currentField != null) {
```
andyHa: not all parents are joined

PR #75, opened 2018-02-15
Default values

Tobsucht:
Properly uses default values defined for a property when using transformValue()
Adds logging to the schema tool
Fixes a NPE in Boolean property
Fixes the definition of default values in Create Table commands (quotation)
Adds tests


src/main/java/sirius/db/mixing/properties/AmountProperty.java
```diff
@@ -72,10 +73,14 @@ public void create(EntityDescriptor descriptor,
 
     @Override
     public Object transformValue(Value value) {
-        if (!value.isFilled()) {
-            return Amount.NOTHING;
-        } else {
+        if (value.isFilled()) {
             return NLS.parseUserString(Amount.class, value.asString());
+        } else {
```
tbiScireum: des if-else ist unnötig komplex in den transformValues

src/main/java/sirius/db/mixing/properties/BooleanProperty.java
```diff
@@ -8,12 +8,14 @@
 
 package sirius.db.mixing.properties;
 
+import com.sun.org.apache.xpath.internal.operations.Bool;
```
fkrone: Not needed import

PR #96, opened 2018-07-04
Fix bug for LocalDate Fields in ES, add computeAggregations, add convenience method

Tobsucht:


src/main/java/sirius/db/mixing/properties/BaseMapProperty.java
```diff
@@ -25,6 +25,16 @@
  */
 public abstract class BaseMapProperty extends Property {
 
+    /**
+     * Contains the name of the field used to store the map key
+     */
+    public static final String KEY = "key";
```
andyHa: isn't that implementation specific to ES? see ESStringMapProperty.KEY ...

PR #100, opened 2018-07-06
Creates only one JedisPool, even if in the beginning multiple Threads…

D3nnisH:
… try to access Redis at the same time


src/main/java/sirius/db/redis/Redis.java
```diff
@@ -219,7 +219,7 @@ private Jedis getConnection() {
         return setupConnection();
     }
 
-    private Jedis setupConnection() {
+    private synchronized Jedis setupConnection() {
```
andyHa: add an else if (sentinelPool != null) {
return sentinelPool.getResource();
} in line 230 - to be 100% race-condition free

PR #106, opened 2018-07-13
Bulking Elastic: get failed requests

Tobsucht:


src/main/java/sirius/db/es/LowLevelClient.java
```diff
@@ -294,7 +294,7 @@ public JSONObject count(List<String> indices, String type, String routing, JSONO
     @Explain("Due to method overloading the compiler cannot deduce which method to pick")
     public JSONObject bulk(List<JSONObject> bulkData) {
         return performPost().rawData(bulkData.stream().map(obj -> obj.toJSONString()).collect(Collectors.joining("\n"))
-                                     + "\n", "application/x-ndjson").execute("_bulk").response();
+                                     + "\n").execute("_bulk").response();
```
Tobsucht: When sending this content-type, the server responds with something like "content-type" is not set.
adapted it as seen in this PR https://github.com/dadoonet/fscrawler/pull/384/files

src/main/java/sirius/db/es/BulkContext.java
```diff
@@ -104,21 +120,28 @@ public BulkContext forceDelete(ElasticEntity entity) {
     }
 
     private void update(ElasticEntity entity, boolean force) {
+        isUpdateRequest = true;
```
andyHa: deletes and updates can be mixed - shouldn't we really rely on this. I'd rather not invoke any afterXXX calls (and document this behaviour) - a bulk delete with afterDeleteCalls wouldn`t be much of a performance win anyway..

Also storing all entities in another list might add some heap pressure (unexpected)

PR #110, opened 2018-07-20
Generate failure message at commit, provide getter for failure message

Tobsucht:


src/main/java/sirius/db/es/BulkContext.java
```diff
@@ -206,7 +206,8 @@ public boolean commit() {
             this.response = response;
             boolean hasErrors = response.getBooleanValue("errors");
 
-            if (hasErrors && Strings.isFilled(failureMessage)) {
+            if (hasErrors) {
+                this.getFailedIds();
                 Exceptions.handle().withSystemErrorMessage(failureMessage).handle();
```
andyHa: getFailureMessage() - don't rely on magic side effects

src/main/java/sirius/db/es/BulkContext.java
```diff
@@ -206,7 +206,8 @@ public boolean commit() {
             this.response = response;
             boolean hasErrors = response.getBooleanValue("errors");
 
-            if (hasErrors && Strings.isFilled(failureMessage)) {
+            if (hasErrors) {
+                this.getFailedIds();
```
andyHa: -

PR #114, opened 2018-08-17
Corrects batch update and delete query for versioned entities

tbiScireum:


src/main/java/sirius/db/jdbc/batch/DeleteQuery.java
```diff
@@ -118,7 +118,7 @@ protected void buildSQL() throws SQLException {
 
         if (descriptor.isVersioned()) {
             if (mf.successiveCall()) {
```
andyHa: sql.append(" ")

src/main/java/sirius/db/jdbc/batch/UpdateQuery.java
```diff
@@ -197,7 +197,7 @@ protected void buildSQL() throws SQLException {
 
         if (descriptor.isVersioned()) {
```
andyHa: sql.append(" ")

PR #120, opened 2018-09-05
Change filled/notFilled implementation for mongo and elastic

Tobsucht:
Filled/NotFilled within mongo and elastic is not only defined by the presence/absence of a field but also whether the
field value is NULL.

### DIFF
--- a/src/main/java/sirius/db/mongo/constraints/MongoFilterFactory.java
+++ b/src/main/java/sirius/db/mongo/constraints/MongoFilterFactory.java
@@ -21,6 +21,7 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.time.ZoneId;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mongo/constraints/MongoFilterFactory.java
```diff
@@ -74,12 +75,14 @@ protected MongoConstraint ltValue(Mapping field, Object value, boolean orEqual)
...
-        return new MongoConstraint(field.toString(), new Document("$exists", true));
+        return effectiveAnd(Arrays.asList(new MongoConstraint(field.toString(), new Document("$exists", true)),
+                                          neValue(field, null)));
     }
 
     @Override
     public MongoConstraint notFilled(Mapping field) {
-        return new MongoConstraint(field.toString(), new Document("$exists", false));
+        return effectiveOr(Arrays.asList(new MongoConstraint(field.toString(), new Document("$exists", false)),
+                                         eqValue(field, null)));
```
D3nnisH: eqValue(field, null) matcht auch "nicht vorhanden", also kannst dir den $exists an der Stelle sparen

PR #124, opened 2018-09-10
Fix reading integer and string int property from MongoDB

Tobsucht:


src/main/java/sirius/db/mixing/properties/IntegerProperty.java
```diff
@@ -91,14 +95,8 @@ public void contributeToTable(Table table) {
...
         description.put(IndexMappings.MAPPING_TYPE, "integer");
-        transferOption(IndexMappings.MAPPING_STORED,
-                       getAnnotation(IndexMode.class),
-                       IndexMode::stored,
-                       description);
-        transferOption(IndexMappings.MAPPING_INDEX,
-                       getAnnotation(IndexMode.class),
-                       IndexMode::indexed,
-                       description);
+        transferOption(IndexMappings.MAPPING_STORED, getAnnotation(IndexMode.class), IndexMode::stored, description);
```
andyHa: please check code format settings

PR #132, opened 2018-10-01
Uses null-safe equal operator

tbiScireum:


src/main/java/sirius/db/jdbc/batch/BatchQuery.java
```diff
@@ -258,4 +260,28 @@ public String toString() {
...
+
+    protected void buildWhere(StringBuilder sql, boolean addVersionConstraint) {
+        sql.append(" WHERE ");
+        Monoflop mf = Monoflop.create();
+        for (Property p : getProperties()) {
+            if (mf.successiveCall()) {
+                sql.append(" AND ");
+            }
+            sql.append(p.getPropertyName());
+            sql.append(" <=> ?");
```
andyHa: use a special Capability and set it true for mysql

PR #133, opened 2018-10-09
Does not reset skip to 0 anymore in Finder#limit(int)

D3nnisH:


src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -138,7 +138,6 @@ public Finder limit(int skip, int limit) {
      * @return the builder itself for fluent method calls
      */
     public Finder limit(int limit) {
-        this.skip = 0;
```
andyHa: plz fix javadoc above

PR #137, opened 2018-10-11
Adds support for StringBoolean Maps to Elastic

Tobsucht:


src/main/java/sirius/db/mixing/properties/StringBooleanMapProperty.java
```diff
@@ -8,29 +8,41 @@
...
 
 /**
  * Represents an {@link StringBooleanMap} field within a {@link Mixable}.
  */
 public class StringBooleanMapProperty extends BaseMapProperty {
 
+    /**
+     * Contains the name of the field used to store the map key
+     */
+    public static final String KEY = "key";
```
andyHa: move to basemap as this is duplicate code vs StringMap

andyHa: or simplie reuse constants as done in StringLocalDateTimeMap

src/main/java/sirius/db/mixing/properties/StringBooleanMapProperty.java
```diff
@@ -8,29 +8,41 @@
...
+import java.util.Collection;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 /**
  * Represents an {@link StringBooleanMap} field within a {@link Mixable}.
  */
 public class StringBooleanMapProperty extends BaseMapProperty {
```
andyHa: should most probably implement ESPropertyInfo?

PR #141, opened 2018-10-22
Adds support to use ":" within token-values

Tobsucht:
Problem was that the compiler interpreted the ":" within the value as the start of a operator.


Tobsucht:
build

Tobsucht:
build

PR #149, opened 2018-11-09
Adds support for elasticsearch suggesters

jmuscireum:
Different to sirius-search, suggesters can be added to any query. This allows to fetch suggestions without the need for additional requests.

### DIFF
--- a/src/main/java/sirius/db/es/ElasticQuery.java
+++ b/src/main/java/sirius/db/es/ElasticQuery.java
@@ -838,12 +838,17 @@ public class ElasticQuery<E extends ElasticEntity> extends Query<ElasticQuery<E>
                     client.search(indices, elastic.determineTypeName(descriptor), routing, skip, limit, buildPayload());
         }
 
-        return response.getJSONObject(KEY_SUGGEST)
-                       .getJSONArray(name)
-                       .stream()
-                       .map(part -> (JSONObject) part)
-                       .map(SuggestPart::makeSuggestPart)
-                       .collect(Collectors.toList());
+        JSONObject responseSuggestions = response.getJSONObject(KEY_SUGGEST);
+
+        if (responseSuggestions == null) {
+            return Collections.emptyList();
+        }
+
+        return responseSuggestions.getJSONArray(name)
+                                  .stream()
+                                  .map(part -> (JSONObject) part)
+                                  .map(SuggestPart::makeSuggestPart)
+                                  .collect(Collectors.toList());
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -483,6 +489,34 @@ public InnerHitsBuilder addCollapsedInnerHits(String name, int size) {
...
+    }
+
+    /**
+     * Adds a suggester.
+     *
+     * @param suggestBuilder a suggest builder
+     * @return the query itself for fluent method calls
+     */
+    public ElasticQuery<E> suggest(SuggestBuilder suggestBuilder) {
+        if (this.suggesters == null) {
```
fkrone: call the suggest method here to prevent duplicated code

src/main/java/sirius/db/es/suggest/SuggestBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Sets the query string to generate suggestions for and the field to generate the suggestions from
+     *
+     * @param field the field to get the suggestions from
+     * @param text  the query to generate suggestions for
+     * @return the builder itself for fluent method calls
+     */
+    public SuggestBuilder on(Mapping field, String text) {
+        this.text = text;
+        this.body.put(PARAM_FIELD, field.getName());
```
fkrone: addBodyParameter call

src/main/java/sirius/db/es/suggest/SuggestBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Creates a new suggest builder.
+     *
+     * @param type the type of the suggester.
+     * @param name the name of the suggester
+     * @return the builder itself for fluent method calls
+     */
+    public static SuggestBuilder create(String type, String name) {
```
andyHa: why not use a constructor?

src/main/java/sirius/db/es/suggest/SuggestPart.java
```diff
@@ -0,0 +1,111 @@
...
+        return length;
+    }
+
+    /**
+     * Returns the options given for this part.
+     *
+     * @return a list of {@link SuggestOption}s
+     */
+    public List<SuggestOption> getOptions() {
+        return options;
```
andyHa: this might return null which is highly discouraged

PR #151, opened 2018-11-15
Jmu/elastic additions

jmuscireum:


src/main/java/sirius/db/es/AggregationBuilder.java
```diff
@@ -20,9 +21,17 @@
  */
 public class AggregationBuilder {
 
-    private static final String NESTED = "nested";
+    public static final String NESTED = "nested";
```
andyHa: javadoc or protected/private

PR #158, opened 2018-11-22
Allows to customize the query compilation in a more fine grained fashion

Tobsucht:
E.g. parseOperation() can be called for non-property tag values.


src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -199,11 +199,20 @@ private C parseExpression() {
...
-                token = treatOperatorAsTokenPart(token);
-            }
+
+            return compileContraint(property, token, skipped);
+        }
+
+        return compileDefaultSearch(searchFields, token);
+    }
+
+    protected C compileContraint(Property property, FieldValue token, boolean skipped) {
```
welzfabian: compileCon**s**traint?

PR #163, opened 2019-01-07
Replaces "-token" with NOT "token" in ES Queries

meg1502:
Fixes: OX-4604


meg1502:
for all compilers not just ES

PR #167, opened 2019-01-21
Does not throw an exception if the result of a mongo aggregation is e…

D3nnisH:
…mpty

### DIFF
--- a/src/main/java/sirius/db/mongo/Finder.java
+++ b/src/main/java/sirius/db/mongo/Finder.java
@@ -10,8 +10,8 @@ package sirius.db.mongo;
 
 import com.google.common.collect.ImmutableList;
 import com.mongodb.BasicDBObject;
-import com.mongodb.client.AggregateIterable;
 import com.mongodb.client.FindIterable;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -381,12 +382,17 @@ public Value aggregateIn(@Nonnull String collection, @Nonnull Mapping field, @No
...
-                                 .first()
-                                 .get("result"));
+            AggregateIterable<Document> queryResult = mongo.db()
+                                                           .getCollection(collection)
+                                                           .aggregate(ImmutableList.of(new BasicDBObject("$match",
+                                                                                                         filterObject),
+                                                                                       new BasicDBObject("$group",
+                                                                                                         groupStage)));
+            if (queryResult.iterator().hasNext()) {
+                return Value.of(queryResult.first().get("result"));
```
andyHa: won't this execute the query twice?

D3nnisH: indeed

PR #173, opened 2019-02-18
Adds helper method which returns currently active indexes for the 'ac…

Tobsucht:
…tive alias'

### DIFF
--- a/src/main/java/sirius/db/es/LowLevelClient.java
+++ b/src/main/java/sirius/db/es/LowLevelClient.java
@@ -266,11 +266,11 @@ public class LowLevelClient {
     }
 
     /**
-     * Returns all indexes which hold the {@link Elastic#ACTIVE_ALIAS} for the given {@link EntityDescriptor}.
+     * Returns all indices which hold the {@link Elastic#ACTIVE_ALIAS} for the given {@link EntityDescriptor}.
      *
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/db/es/LowLevelClient.java
```diff
@@ -264,6 +265,19 @@ public boolean aliasExists(String alias) {
         }
     }
 
+    /**
+     * Returns all indexes which hold the {@link Elastic#ACTIVE_ALIAS} for the given {@link EntityDescriptor}.
+     *
+     * @return a list of all indexes which hold the {@link Elastic#ACTIVE_ALIAS} for the given {@link EntityDescriptor}
+     */
+    public List<String> getIndexesForAlias(EntityDescriptor ed) {
```
andyHa:
```suggestion
    public List<String> getIndicesForAlias(EntityDescriptor ed) {
```

src/main/java/sirius/db/es/LowLevelClient.java
```diff
@@ -264,6 +265,19 @@ public boolean aliasExists(String alias) {
         }
     }
 
+    /**
+     * Returns all indexes which hold the {@link Elastic#ACTIVE_ALIAS} for the given {@link EntityDescriptor}.
```
andyHa:
```suggestion
     * Returns all indices which hold the {@link Elastic#ACTIVE_ALIAS} for the given {@link EntityDescriptor}.
```

PR #175, opened 2019-02-18
Limits the log entry of "slow queries" as well as debug logs.

andyHa:
This is essential when logging bulk inserts as otherwise, the
generated log messages will be very large.

Fixes: SIRI.24


src/main/java/sirius/db/es/RequestBuilder.java
```diff
@@ -45,6 +46,7 @@
     private static final String PARAM_ROUTING = "routing";
     private static final String PARAM_VERSION = "version";
     private static final String PARAM_ERROR = "error";
+    private static final int MAX_CONTET_LONG_LENGTH = 256;
```
D3nnisH: `MAX_CONTENT_LONG_LENGTH`

PR #209, opened 2019-06-06
Make StringLists usable for SQLEntities

welzfabian:


src/main/java/sirius/db/jdbc/Capability.java
```diff
@@ -43,6 +43,11 @@
      */
     NULL_SAFE_OPERATOR,
 
+    /**
+     * Signals that the database supports lists.
+     */
+    LISTS,
+
     GENERATED_KEYS;
```
andyHa: add javadoc here

src/main/java/sirius/db/mixing/properties/StringListProperty.java
```diff
@@ -113,12 +122,18 @@ protected Object transformFromMongo(Value object) {
 
     @Override
     protected Object transformFromJDBC(Value object) {
-        return Arrays.asList(object.coerce(String[].class, EMPTY_STRING_ARRAY));
+        if (oma.getDatabase(descriptor.getRealm()).hasCapability(Capability.LISTS)) {
```
andyHa: could we please cache this flag?

src/main/java/sirius/db/mixing/properties/StringListProperty.java
```diff
@@ -113,12 +122,18 @@ protected Object transformFromMongo(Value object) {
...
+            return Arrays.asList(object.coerce(String[].class, EMPTY_STRING_ARRAY));
+        }
+        return Arrays.stream(object.asString().split(",")).filter(Strings::isFilled).collect(Collectors.toList());
     }
 
     @Override
     protected Object transformToJDBC(Object object) {
-        return object;
+        if (oma.getDatabase(descriptor.getRealm()).hasCapability(Capability.LISTS)) {
```
andyHa: s.a.

src/main/java/sirius/db/mixing/properties/StringListProperty.java
andyHa: maybe even call this in the constructor? or a verify / link run

andyHa: maybe also extend the description of the class javadoc of this class an StringList to state that it can be used with JDBC databases, how it is handled and what limits are imposed.

src/main/java/sirius/db/jdbc/Capability.java
```diff
@@ -43,6 +43,11 @@
      */
     NULL_SAFE_OPERATOR,
 
+    /**
+     * Signals that the database supports lists.
```
andyHa:
```suggestion
     * Signals that the database supports storing lists as field values.
```

src/main/java/sirius/db/jdbc/schema/BasicDatabaseDialect.java
```diff
@@ -383,6 +383,9 @@ protected String getTypeName(TableColumn column) {
         if (Types.BLOB == type || Types.VARBINARY == type || Types.LONGVARBINARY == type) {
             return "LONGBLOB";
         }
+        if (Types.ARRAY == type) {
```
andyHa: shouldn't we fix this in the string list property - this seems out of place..

src/test/java/sirius/db/jdbc/SQLStringListPropertyEntitySpec.groovy
```diff
@@ -0,0 +1,55 @@
...
+
+import sirius.kernel.BaseSpecification
+import sirius.kernel.di.std.Part
+
+class SQLStringListPropertyEntitySpec extends BaseSpecification {
+
+    @Part
+    static OMA oma
+
+    def "test many list entries"() {
```
andyHa: add a test with a short string field and many list items so that data truncation happens and ensure that a proper and understandable exception is shown / logged

src/main/java/sirius/db/mixing/properties/StringListProperty.java
welzfabian: could not be executed in constructor or link, because OMA was not ready then

PR #211, opened 2019-06-07
Aha/complex delete

andyHa:


src/main/java/sirius/db/mixing/properties/BaseEntityRefListProperty.java
```diff
@@ -182,6 +194,20 @@ public void link() {
                                 + " The only supported DeleteHandler is IGNORE!.", this, getDescriptor());
                 return;
             }
+
+            // If a cascade delete handler is present and the referenced entity is not explicitely marked as
+            // "non complex" and we're within the IDE or running as a test, we force the system to compute / lookup
+            // the associated NLS keys which might be required to generated appropriate deletion logs or rejection
+            // errors. (Otherwise this might be missed while developing or testing the system..)
+            if (getReferencedDescriptor().getAnnotation(ComplexDelete.class).map(ComplexDelete::value).orElse(true)
```
sabieber: is this the right default, shouldnt this be `orElse(false)`?

src/main/java/sirius/db/mixing/properties/BaseEntityRefProperty.java
```diff
@@ -180,6 +192,20 @@ public void link() {
                                 + "The only supported DeleteHandler is IGNORE!.", this, getDescriptor());
                 return;
             }
+
+            // If a cascade delete handler is present and the referenced entity is not explicitely marked as
+            // "non complex" and we're within the IDE or running as a test, we force the system to compute / lookup
+            // the associated NLS keys which might be required to generated appropriate deletion logs or rejection
+            // errors. (Otherwise this might be missed while developing or testing the system..)
+            if (getReferencedDescriptor().getAnnotation(ComplexDelete.class).map(ComplexDelete::value).orElse(true)
```
sabieber: is this the right default, shouldnt this be `orElse(false)`?

src/main/java/sirius/db/mixing/properties/BaseEntityRefListProperty.java
andyHa: No, if an entity performs cascading deletes, we consider that as complex unless noted otherwise.

src/main/java/sirius/db/mixing/properties/BaseEntityRefProperty.java
andyHa: s.a.

src/main/java/sirius/db/mixing/Property.java
```diff
@@ -136,13 +136,15 @@ protected Property(@Nonnull EntityDescriptor descriptor, @Nonnull AccessPath acc
         this.descriptor = descriptor;
         this.accessPath = accessPath;
         this.field = field;
-        this.propertyKey = field.getDeclaringClass().getSimpleName() + "." + field.getName();
+        this.propertyKey = EntityDescriptor.determineTranslationSource(field.getDeclaringClass()).getSimpleName()
```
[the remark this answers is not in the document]
andyHa: because the field class may be != the entity class (e.g. a composite or a mixin.)

src/main/java/sirius/db/mixing/properties/BaseEntityRefListProperty.java
sabieber: aahhhhh now i get it. thx

PR #232, opened 2019-09-05
Don't use FixedString for enums in Clickhouse by default

jmuscireum:
Instead String is used. This is necessary, because the length can't be easily changed after the table is created. If the length is set via `@Length` fall back to FixedString in order to be backward compatible.


src/main/java/sirius/db/jdbc/schema/ClickhouseDatabaseDialect.java
```diff
@@ -116,7 +118,22 @@ private String getClickHouseType(TableColumn column) {
                                                          column.getSource()));
     }
 
+    /**
+     * Determines the type for string columns.
+     * <p>
+     * {@link EnumProperty enum properties} should use the type <tt>String</tt>. Only if the length is explicitly
```
andyHa:
```suggestion
     * {@link EnumProperty Enum properties} should use the type <tt>String</tt>. Only if the length is explicitly
```

PR #237, opened 2019-09-19
Moves the logic to generate Mongo ids into MongoEntity

meg1502:
So overriding the id generation can be done easier


src/main/java/sirius/db/mongo/MongoEntity.java
```diff
@@ -64,4 +68,8 @@ public String getId() {
     protected void setId(String id) {
         this.id = id;
     }
+
+    protected String generateId() {
```
andyHa: JavaDoc

PR #245, opened 2019-10-14
Aha/improved updates

andyHa:


src/main/java/sirius/db/jdbc/UpdateStatement.java
```diff
@@ -163,26 +181,41 @@ public UpdateStatement setToToday(Mapping field) {
...
      * @param value the value to enforce
      * @return the query itself for fluent method calls
      */
-    public UpdateStatement where(Mapping field, Object value) {
+    public UpdateStatement where(Mapping field, Operator op, Object value) {
         prepareWhere();
         append(determineEffectiveColumnName(field));
-        if (value == null) {
+        if (value == null && op == Operator.EQ) {
             append(" IS NULL");
```
michaelhaufler: Wie ist denn das bei value null und op==Operator.NE

src/main/java/sirius/db/jdbc/UpdateStatement.java
```diff
@@ -163,26 +181,41 @@ public UpdateStatement setToToday(Mapping field) {
...
+     * @param op    the operator to use
      * @param value the value to enforce
      * @return the query itself for fluent method calls
      */
-    public UpdateStatement where(Mapping field, Object value) {
+    public UpdateStatement where(Mapping field, Operator op, Object value) {
         prepareWhere();
         append(determineEffectiveColumnName(field));
-        if (value == null) {
+        if (value == null && op == Operator.EQ) {
```
idlira: Not Equal condition should also be evaluated as "IS NOT NULL" (from my knowledge of Oracle DBs)

... else if (value == null && op == Operator.NE) {
   append(" IS NOT NULL");

PR #248, opened 2019-10-21
Prevents a misuse of the BatchContext which could drain the connectio…

andyHa:
…n pool.

If a query was created but accessed after the pool has been closed, a connection
was opened but never closed. We now throw an IllegalStateException
(which might get wrapped into a HandledException) to display and aport this
condition (without leaking a connection).


src/main/java/sirius/db/jdbc/batch/BatchContext.java
```diff
@@ -99,10 +114,10 @@ protected void safeClose() {
 
             query.safeClose();
         }
-        queries.clear();
+//        queries = null;
```
tbiScireum: soll das so drin bleiben ?

PR #256, opened 2019-11-25
Properly logs a warning when linking a reference property runs into a…

andyHa:
…n error.

This commonly happens if an entity from a disabled framework is referenced.
We now output the property and descriptor of the entity which caused the error
instead of just outputting the name of the missing entity type.

### DIFF
--- a/src/main/java/sirius/db/mixing/properties/BaseEntityRefProperty.java
+++ b/src/main/java/sirius/db/mixing/properties/BaseEntityRefProperty.java
@@ -215,6 +215,10 @@ public abstract class BaseEntityRefProperty<I, E extends BaseEntity<I>, R extend
                     Mixing.LOG.WARN("Error in property %s of %s: The field is not marked as NullAllowed,"
                                     + " therefore SET_NULL is not a valid delete handler!", this, getDescriptor());
                 }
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mixing/properties/BaseEntityRefProperty.java
```diff
@@ -198,29 +198,34 @@ protected void onBeforeSaveChecks(Object entity) {
...
 
-        if (deleteHandler == BaseEntityRef.OnDelete.CASCADE) {
-            getReferencedDescriptor().addCascadeDeleteHandler(this::onDeleteCascade);
-        } else if (deleteHandler == BaseEntityRef.OnDelete.SET_NULL) {
-            if (!isNullable()) {
-                Mixing.LOG.WARN("Error in property %s of %s. The field is not marked as NullAllowed,"
-                                + " therefore SET_NULL is not a valid delete handler!", this, getDescriptor());
+                ensureLabelsArePresent(this, referencedDescriptor, deleteHandler);
             }
-            if (entityRef.hasWriteOnceSemantics()) {
```
jmuscireum: Was this intentional?

PR #274, opened 2020-02-07
BREAKING: Specifies functional interface as indicated by sonarqube

meg1502:


src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -37,7 +37,7 @@
     /**
      * Represents a value parsed for a field.
      */
-    protected static class FieldValue {
+    protected static class  FieldValue {
```
andyHa: meep

meg1502:
```suggestion
    protected static class FieldValue {
```

PR #279, opened 2020-02-13
Introduces support for MongoDB Multi-Language strings

idlira:
Introduces a new `MultiLanguageString` property type for usage by MongoEntity objects, allowing to store multiple translations of a String und list containing sub-elements with the `lang` and `text` properties.

Refer to the MongoMultiLanguageStringPropertySpec for a detailed usage of all exposed methods.


src/main/resources/db_de.properties
```diff
@@ -24,6 +24,8 @@ DateRange.thisWeek = Diese Woche
 DateRange.thisYear = Dieses Jahr
 DateRange.today = Heute
 DateRange.yesterday = Gestern
+MultiLanguageString.invalidLanguage = Ungültiger Wert '${language}' als Sprachecode.
```
ymo-sci:
```suggestion
MultiLanguageString.invalidLanguage = Ungültiger Wert '${language}' als Sprachcode.
```

src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    protected Object transformToMongo(Object object) {
+        if (object instanceof List) {
+            return object;
+        }
+
+        List<Document> texts = new ArrayList<>();
+        ((Map<String, String>) object).forEach((language, text) -> {
```
andyHa: .filter(Strings::isFilled) ?

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public boolean hasText(String language) {
+        return data().containsKey(language);
+    }
+
+    /**
+     * Returns the text associated with the default language.
+     *
+     * @return the text
+     */
```
andyHa: @throws HandledException if no transaltion is available for the <tt>defaultLanguage</tt>

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+    public String getRequiredText() {
+        return getRequiredText(defaultLanguage);
+    }
+
+    /**
+     * Returns the text associated with a given language.
+     *
+     * @param language the language code
+     * @return the text or an {@link sirius.kernel.health.HandledException} if not existent
+     */
```
andyHa: @throws ...s.a.

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+    public String getRequiredText() {
+        return getRequiredText(defaultLanguage);
+    }
+
+    /**
+     * Returns the text associated with a given language.
+     *
+     * @param language the language code
+     * @return the text or an {@link sirius.kernel.health.HandledException} if not existent
```
andyHa: put exception into throws

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+    public boolean hasText(String language) {
+        return data().containsKey(language);
+    }
+
+    /**
+     * Returns the text associated with the default language.
+     *
+     * @return the text
+     */
+    public String getRequiredText() {
```
andyHa: @Nonnull

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+        return getRequiredText(defaultLanguage);
+    }
+
+    /**
+     * Returns the text associated with a given language.
+     *
+     * @param language the language code
+     * @return the text or an {@link sirius.kernel.health.HandledException} if not existent
+     */
+    public String getRequiredText(String language) {
```
andyHa: @Nonnull

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        return fetchText(language);
+    }
+
+    /**
+     * Returns an optional text associated with the default language.
+     *
+     * @return an Optional String containing the text, otherwise an empty Optional
+     */
+    public Optional<String> getText() {
```
andyHa: @Nonnull

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+        return getText(defaultLanguage);
+    }
+
+    /**
+     * Returns an optional text associated with a given language.
+     *
+     * @param language the language code
+     * @return an Optional String containing the text, otherwise an empty Optional
+     */
+    public Optional<String> getText(String language) {
```
andyHa: @Nonnull

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        return Optional.of(fetchText(language));
+    }
+
+    /**
+     * Returns the text associated with the default language.
+     *
+     * @return the text if it exists, otherwise <tt>null</tt>
+     */
+    public String fetchText() {
```
andyHa: @Nullable

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+     *
+     * @param language         the language code
+     * @param fallbackLanguage the alternative language code
+     * @return the text found under <tt>language</tt>, if none found under <tt>fallbackLanguage</tt>, <tt>null</tt> otherwise
+     */
+    public String fetchText(String language, String fallbackLanguage) {
+        return data().getOrDefault(language, data().getOrDefault(fallbackLanguage, fetchText(defaultLanguage)));
+    }
+
+    private void assertLanguage(String language) {
```
andyHa: assertValidLanguage (?)

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+                            .set("language", language)
+                            .handle();
+        }
+    }
+
+    public String getDefaultLanguage() {
+        return defaultLanguage;
+    }
+
+    public void setDefaultLanguage(String defaultLanguage) {
```
andyHa: is there a good reason to change this?

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+ * <p>
+ * These are stored in MongoDB as an array containing sub-documents
+ * containing a <tt>lang</tt> and <tt>text</tt> properties
+ *
+ * @see sirius.db.mixing.properties.MultiLanguageStringProperty
+ */
+public class MultiLanguageString extends SafeMap<String, String> {
+
+    private static final List<String> supportedLanguages =
+            Sirius.getSettings().getConfig("mongo").getStringList("supportedLanguages");
```
andyHa: this static initializer is evaluated too early (config might not be present yet)
Load on first use(?) Also use a Set in favor of a list(?)

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+
+    private String defaultLanguage;
```
andyHa: do we really need this or would we rather use NLS.getCurrentLanguage below?

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+     * @return the text found under <tt>language</tt>, if none found under <tt>fallbackLanguage</tt>, <tt>null</tt> otherwise
+     */
+    public String fetchText(String language, String fallbackLanguage) {
+        return data().getOrDefault(language, data().getOrDefault(fallbackLanguage, fetchText(defaultLanguage)));
+    }
+
+    private void assertLanguage(String language) {
+        if (!supportedLanguages.contains(language)) {
+            throw Exceptions.createHandled()
+                            .withNLSKey("MultiLanguageString.invalidLanguage")
```
andyHa: this is something the programmer screws up, not the user. Therefore it would be totally fine to throw an untranslated system error in english (no love for coders around here :-P ) - still no need to change this, just a hint for the future

src/main/resources/db_de.properties
```diff
@@ -24,6 +24,8 @@ DateRange.thisWeek = Diese Woche
 DateRange.thisYear = Dieses Jahr
 DateRange.today = Heute
 DateRange.yesterday = Gestern
+MultiLanguageString.invalidLanguage = Ungültiger Wert '${language}' als Sprachcode.
+MultiLanguageString.textDoesNotExist = Für den Sprachcode '${language}' ist kein Text definiert.
```
andyHa: this message will show up somewhere and we'll hate ourselfes, as we don't know the _field_ :-P - but I also don't have a quick solution for this other than passing in the field name or Property or Entity+Mapping to determine it :-/

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
idlira: Do we want to annotate all `fetchText` methods with a @nullable?
If so, basically any method that do not return void needs either nullable or nonnull

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
idlira: It's an attempt to make the framework as generic as possible.
I'm thinking about a scenario where the default language could derive from the data context, like the assortment's default language, or a language picker int the UI, which could differ from the user's frontend display language.

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
idlira: It's an attempt to make the framework as generic as possible.
I'm thinking about a scenario where the default language could derive from the data context, like the assortment's default language, or a language picker int the UI, which could differ from the user's frontend display language.

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
idlira: Not very sure, as this can be data-relevant. If we suddenly get assortments in an unexpected language, is not really the programmer's fault.

src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
```diff
@@ -0,0 +1,96 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+/**
+ * Represents an {@link MultiLanguageString} field in a {@link sirius.db.mongo.MongoEntity}.
```
meg1502:
```suggestion
 * Represents a {@link MultiLanguageString} field in a {@link sirius.db.mongo.MongoEntity}.
```

src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
```diff
@@ -0,0 +1,96 @@
...
+ * <p>
+ * Multi-Language Strings are stored as a list of nested objects which contain a <tt>lang</tt> and
+ * a <tt>text</tt> properties.
```
meg1502:
```suggestion
 * a <tt>text</tt> property.
```

not 100% sure

src/main/resources/db_de.properties
idlira: Thought about it too, but also did not find a quick elegant way to solve it. This property is new, we can still force to pass the property and/or Mapping, but would then be inconsistent with the other properties (and extending it to the props would be quite a breaking change if we enforce it)

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+import sirius.kernel.health.Exceptions;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Provides a language-text map as property value.
+ * <p>
+ * These are stored in MongoDB as an array containing sub-documents
+ * containing a <tt>lang</tt> and <tt>text</tt> properties
```
meg1502:
```suggestion
 * containing a <tt>lang</tt> and a <tt>text</tt> property.
```

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @see sirius.db.mixing.properties.MultiLanguageStringProperty
+ */
+public class MultiLanguageString extends SafeMap<String, String> {
```
meg1502: I'm kinda missing a 'getText' method without parameters that returns NLS.getCurrentLanguage and falls back to the default language

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
meg1502: shouldnt this maybe use NLS.currentLang with a fallback to defaultLang instead of always using the defaultLang?

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
andyHa: we could argue about the nonnull for optionals as it is kind of a no-brainer
in this case i'd welcome the @nullable as then the compiler can enforce some checks

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
andyHa: I'd personally make  the framework as simple as possible and stick with defaultLanguage == NLS.getCurrentLanguage - everything else should be handled one level higher(?) - IMHO

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
andyHa: right +1

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
andyHa: i'd also like this in favor of defaultLanguage

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
andyHa: +1

src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
idlira: The document has 2 properties: "lang" and "text". But will improve the JavaDoc so they are in the same line

src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
idlira: Store empty string, but reject null

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
idlira: Got rid of defaultLanguage. Using NLS.getCurrentLanguage instead.

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
idlira: NLS.getCurrentLanguage

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
idlira: Using a Set, lazy-load

PR #280, opened 2020-02-13
SIRI-172 - add support for database supported optimistic locking patterns

andyHa:


src/main/java/sirius/db/mixing/IntegrityConstraintFailedException.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Creates a new instance without any reference to another error.
+     */
+    public IntegrityConstraintFailedException() {
+    }
+
+    /**
+     * Creates a new instance with a reference to the given cause.
+     *
+     * @param cause   the cuase of this error
```
meg1502:
```suggestion
     * @param cause   the cause of this error
```

PR #281, opened 2020-02-13
Builds the routing table independent of the mapping generation.

andyHa:
Otherwise, nodes which do not update the mappings, will have no
routing information and generate invalid data and queries.

Fixes: SIRI-173

### DIFF
--- a/src/main/java/sirius/db/es/IndexMappings.java
+++ b/src/main/java/sirius/db/es/IndexMappings.java
@@ -92,8 +92,13 @@ public class IndexMappings implements Startable {
             return;
         }
 
+        computeRoutingTable();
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/db/es/IndexMappings.java
```diff
@@ -92,10 +92,15 @@ public void started() {
             return;
         }
 
+        computeRoutingTable();
         checkAndUpdateIndices();
         elastic.readyFuture.success();
     }
 
+    private void computeRoutingTable() {
+        mixing.getDesciptors().forEach(this::determineRouting);
```
idlira: not descriptors?

PR #285, opened 2020-02-19
Fixes the AmountProperty initialization check

idlira:
The previous check was not working when fields of type Amount were used inside Composites.

Fixes: SIRI-176

### DIFF
--- a/src/main/java/sirius/db/mixing/properties/AmountProperty.java
+++ b/src/main/java/sirius/db/mixing/properties/AmountProperty.java
@@ -60,7 +60,7 @@ public class AmountProperty extends NumberProperty implements SQLPropertyInfo, E
             AmountProperty amountProperty = new AmountProperty(descriptor, accessPath, field);
 
             try {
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/db/mixing/properties/AmountProperty.java
```diff
@@ -60,7 +60,7 @@ public void create(EntityDescriptor descriptor,
             AmountProperty amountProperty = new AmountProperty(descriptor, accessPath, field);
 
             try {
-                if (field.get(descriptor.getType().getDeclaredConstructor().newInstance()) == null) {
+                if (field.get(accessPath.apply(descriptor.getType().getDeclaredConstructor().newInstance())) == null) {
```
andyHa: add a testcase to reproduce this bug / ensure that it is fixed? :)

idlira: Not sure how to tap into that part of the code in a test case

PR #287, opened 2020-02-21
Fixes a class Cast Exception

tbiScireum:
if the returned value was a long we would try to cast it to a string which would fail with an exception
- Fixes: SE-9343


src/main/java/sirius/db/es/Elastic.java
```diff
@@ -244,7 +244,7 @@ protected String determineRouting(EntityDescriptor ed, ElasticEntity entity) {
             return null;
         }
 
-        return (String) property.getValueForDatasource(Elastic.class, entity);
+        return Value.of(property.getValueForDatasource(Elastic.class, entity)).asString();
```
andyHa: String.valueOf?

PR #288, opened 2020-02-21
Fixes the name of the "routing" parameter in bulk requests.

andyHa:


src/test/java/sirius/db/es/BulkContextSpec.groovy
```diff
@@ -36,6 +36,19 @@ class BulkContextSpec extends BaseSpecification {
         elastic.select(BatchTestEntity.class).count() == 3
     }
 
+    def "batch insert with routing works"() {
+        setup:
+        BulkContext btx = elastic.batch()
```
meg1502: Add ```elastic.select(RoutedBatchTestEntity.class).delete``` to make test more independent ?

src/main/java/sirius/db/mixing/query/BaseQuery.java
```diff
@@ -144,7 +144,7 @@ public void iterateAll(Consumer<E> consumer) {
         if (limit > MAX_LIST_SIZE) {
             throw Exceptions.handle()
                             .to(Mixing.LOG)
-                            .withSystemErrorMessage("When using 'queryList' as most %s items can be selected. "
+                            .withSystemErrorMessage("When using 'queryList' at most %s items can be selected. "
```
idlira:
```suggestion
                            .withSystemErrorMessage("A limit of %s items can be selected when using 'queryList'. "
```

PR #293, opened 2020-02-26
Makes forceEmpty in MongoOneInField work with list fields

meg1502:
As the field is not null, but an empty Array we need to check for either of one,
just as it already the case in orEmpty

This only changes this:
```
oneInField(field, emptyList).forceEmpty().build()  -> $eq: null
```
to
```
oneInField(field, emptyList).forceEmpty().build()  -> $or: $eq null, $eq []
```

oneInFields with filled lists or without forceEmpty are unchanged

### DIFF
--- a/src/main/java/sirius/db/mongo/constraints/MongoOneInField.java
+++ b/src/main/java/sirius/db/mongo/constraints/MongoOneInField.java
@@ -29,13 +29,10 @@ class MongoOneInField extends OneInField<MongoConstraint> {
     public MongoConstraint build() {
         if (values.isEmpty()) {
             if (forceEmpty) {
-                List<MongoConstraint> clauses = new ArrayList<>();
-                clauses.add(factory.notFilled(field));
-                clauses.add(factory.isEmptyArray(field));
 
-                return factory.or(clauses);
+                return factory.or(factory.notFilled(field), factory.isEmptyArray(field));
             }
-
+            
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mongo/constraints/MongoOneInField.java
```diff
@@ -30,7 +29,11 @@ protected MongoOneInField(MongoFilterFactory factory, Mapping field, List<?> val
     public MongoConstraint build() {
         if (values.isEmpty()) {
             if (forceEmpty) {
-                return factory.notFilled(field);
+                List<MongoConstraint> clauses = new ArrayList<>();
+                clauses.add(factory.notFilled(field));
+                clauses.add(factory.isEmptyArray(field));
+
+                return factory.or(clauses);
```
andyHa: factory.or(factory.notFilled(field), factory.isEmptyArray(field)); ?

src/main/java/sirius/db/mongo/constraints/MongoOneInField.java
```diff
@@ -30,9 +29,10 @@ protected MongoOneInField(MongoFilterFactory factory, Mapping field, List<?> val
     public MongoConstraint build() {
         if (values.isEmpty()) {
             if (forceEmpty) {
-                return factory.notFilled(field);
-            }
```
sabieber: is this empty line intended?

PR #295, opened 2020-02-28
Adds 'wasCreated' method into BaseEntity to be called after save

meg1502:
As the correct way to check this at this point is not obvious.
Without this the most obvious but faulty way was to use 'isNew()' here - but this would always return false as the id is already set at this point. The persistedData of the entity is not yet updated so we can check if the ID has been changed.
All provided tests would fail with 'isNew()' in the after save handler.

### DIFF
--- a/src/main/java/sirius/db/jdbc/SQLEntity.java
+++ b/src/main/java/sirius/db/jdbc/SQLEntity.java
@@ -33,13 +33,13 @@ public abstract class SQLEntity extends BaseEntity<Long> {
     protected static OMA oma;
 
     /**
-     * Contains the unique ID of the entity.
+     * Contains the unique {@link #ID} of the entity.
      * <p>
      * This is automatically assigned by the database. Never assign manually a value unless you are totally aware
      * of what you're doing. Also you should not use this ID for external purposes (e.g. as customer number or
      * invoice number) as it cannot be changed.
+     * <p>
      */
-    public static final Mapping ID = Mapping.named("id");
     protected long id = NON_PERSISTENT_ENTITY_ID;
 
     @Transient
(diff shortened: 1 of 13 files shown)


src/main/java/sirius/db/mixing/annotations/AfterSave.java
```diff
@@ -17,6 +18,8 @@
 
 /**
  * Used to mark methods in {@link Mixable}s which will be called once an entity was saved.
```
idlira:
```suggestion
 * Used to mark methods in {@link Mixable mixables} which will be called once an entity was saved.
```

src/main/java/sirius/db/mixing/BaseEntity.java
```diff
@@ -49,6 +49,13 @@
     @Transient
     protected Map<Property, Object> persistedData = Maps.newHashMap();
 
+    /**
+     * Contains the unique id of the entity.
+     * <p>
+     * Subclasses need to create a field with the name {@code id} and the type {@code <I>}.
```
idlira:
```suggestion
     * The actual {@code id} field with corresponding {@code <I>} type is defined by the underlying implementations.
```

src/main/java/sirius/db/jdbc/SQLEntity.java
```diff
@@ -33,15 +33,13 @@
     protected static OMA oma;
 
     /**
-     * Contains the unique ID of the entity.
+     * Contains the unique {@link #ID} of the entity.
      * <p>
      * This is automatically assigned by the database. Never assign manually a value unless you are totally aware
      * of what you're doing. Also you should not use this ID for external purposes (e.g. as customer number or
      * invoice number) as it cannot be changed.
      * <p>
```
jmuscireum:
```suggestion
```

PR #297, opened 2020-03-04
SIRI-188: Suppression of routings for ES

andyHa:


src/main/java/sirius/db/es/Elastic.java
```diff
@@ -434,34 +448,62 @@ public static ContextInfo routedBy(String value) {
...
+    public boolean isRouted(EntityDescriptor entityDescriptor) {
+        return routeTable.containsKey(entityDescriptor);
+    }
+
+    /**
+     * Determines if the usage of the routing for the given descriptor has been suppressed.
+     * <p>
+     * Via the config list {@link #suppressedRoutings} (<tt>elasticsearch.suppressedRoutings</tt>) the usage of
+     * a routing field can entirely be disabled for all listed entities. This might be useful when migrating
+     * form an unrouted to a routed index or even if the routing is only feasible in some scenarios.
```
ymo-sci:
```suggestion
     * from an unrouted to a routed index or even if the routing is only feasible in some scenarios.
```

src/main/java/sirius/db/es/Elastic.java
```diff
@@ -434,24 +448,38 @@ public static ContextInfo routedBy(String value) {
...
+        String routing = context.apply(CONTEXT_ROUTING).getString();
+        if (routing == null && isRouted(entityDescriptor)) {
             LOG.WARN("Trying to FIND an entity of type '%s' with id '%s' without providing a routing! "
                      + "This will most probably return an invalid result!\n%s",
-                     ed.getType().getName(),
+                     entityDescriptor.getType().getName(),
+                     id,
+                     ExecutionPoint.snapshot());
+        } else if (routing != null && !isRouted(entityDescriptor)) {
+            LOG.WARN("Trying to FIND an unrouted entity of type '%s' with id '%s' while a routing! "
```
sabieber:
```suggestion
            LOG.WARN("Trying to FIND an unrouted entity of type '%s' with id '%s' with a routing! "
```

PR #300, opened 2020-03-30
Permits to distinguish the routing suppressions for the read and the …

andyHa:
…write index.

With the previous changes we introduced the possibility to redirect all writes for
and entity into a new index. We now also support to enable and disable the
routing suppression of both indices independent of each other.

This can be achieved using the "es-index" command. Using this facility one can
migrate from an unrouted index to a routed one an vice versa.

### DIFF
--- a/src/main/java/sirius/db/es/ESIndexCommand.java
+++ b/src/main/java/sirius/db/es/ESIndexCommand.java
@@ -124,7 +124,7 @@ public class ESIndexCommand implements Command {
             output.blankLine();
         } else if (args.length() == 4) {
             EnumSet<Elastic.RoutingAccessMode> modes = EnumSet.noneOf(Elastic.RoutingAccessMode.class);
-            if ("Y".equals(args.at(3).asString())) {
+            if ("Y".equals(args.at(2).asString())) {
                 modes.add(Elastic.RoutingAccessMode.READ);
             }
             if ("Y".equals(args.at(4).asString())) {
--- a/src/main/java/sirius/db/es/ESIndexCommand.java
+++ b/src/main/java/sirius/db/es/ESIndexCommand.java
@@ -127,7 +127,7 @@ public class ESIndexCommand implements Command {
             if ("Y".equals(args.at(2).asString())) {
                 modes.add(Elastic.RoutingAccessMode.READ);
             }
-            if ("Y".equals(args.at(4).asString())) {
+            if ("Y".equals(args.at(3).asString())) {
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/db/es/ESIndexCommand.java
```diff
@@ -94,10 +95,52 @@ private boolean executeSubCommand(Output output, Values args) {
...
+            if ("Y".equals(args.at(3).asString())) {
+                modes.add(Elastic.RoutingAccessMode.READ);
+            }
+            if ("Y".equals(args.at(4).asString())) {
+                modes.add(Elastic.RoutingAccessMode.WRITE);
+            }
+            elastic.updateRoutingSuppression(descriptor, modes);
+            output.apply("Suppression of %s has been updated...", descriptor.getType());
+            output.blankLine();
+        }
```
qw3ry: Passen die Indizes hier? Mir kommt das etwas spanisch vor, z.B. `... if (args.length() ==4) { ... args.at(4) ... }` müsste nach meinem Verständnis knallen

src/main/java/sirius/db/es/ESIndexCommand.java
```diff
@@ -94,10 +95,52 @@ private boolean executeSubCommand(Output output, Values args) {
...
+            output.line("Reset suppressions to default:");
+            output.line("es-index suppress-routing <Entity> \"-\"");
+            output.blankLine();
+        } else if ("-".equals(args.at(2).asString())) {
+            elastic.updateRoutingSuppression(descriptor, null);
+            output.apply("Default suppression of %s has been restored...", descriptor.getType());
+            output.blankLine();
+        } else if (args.length() == 4) {
+            EnumSet<Elastic.RoutingAccessMode> modes = EnumSet.noneOf(Elastic.RoutingAccessMode.class);
+            if ("Y".equals(args.at(3).asString())) {
```
jmuscireum:
```suggestion
            if ("Y".equals(args.at(2).asString())) {
```

src/main/java/sirius/db/es/ESIndexCommand.java
```diff
@@ -94,10 +95,52 @@ private boolean executeSubCommand(Output output, Values args) {
...
+                modes.add(Elastic.RoutingAccessMode.READ);
+            }
+            if ("Y".equals(args.at(4).asString())) {
```
jmuscireum:
```suggestion
            if ("Y".equals(args.at(3).asString())) {
```

PR #302, opened 2020-04-04
Generates a proper constraint for list:-

andyHa:


src/main/java/sirius/db/mixing/query/constraints/FilterFactory.java
```diff
@@ -328,6 +328,9 @@ public C lteOrEmpty(Mapping field, @Nullable Object value) {
 
     /**
      * Generates a constraint which ensures that the given field is not filled (null)
+     * <p>
+     * Note that in some cases (e.g. MongoDB) an empty list will not b matched by this
```
sabieber, line 332:
```suggestion
     * Note that in some cases (e.g. MongoDB) an empty list will not be matched by this
```

src/test/java/sirius/db/mongo/MongoQueryCompilerSpec.groovy
```diff
@@ -0,0 +1,47 @@
...
+    private static Mixing mixing
+
+    def "listField:- generates an isEmptyListConstraint"() {
+        when: "We create an example entity with an empty list (superPowers)"
+        MangoTestEntity testEntity = new MangoTestEntity()
+        testEntity.firstname = "Compiler"
+        testEntity.lastname = "Test"
+        and:
+        mango.update(testEntity)
+        and: "And we search for exact that entity while enforcing an empty list."
```
sabieber, line 30:
```suggestion
        and: "And we search for exactly that entity while enforcing an empty list."
```

PR #304, opened 2020-04-08
Adds helper methods to directly obtain the database used to persist an…

andyHa:
… entity type.

### DIFF
--- a/src/main/java/sirius/db/jdbc/OMA.java
+++ b/src/main/java/sirius/db/jdbc/OMA.java
@@ -82,6 +82,20 @@ public class OMA extends BaseMapper<SQLEntity, SQLConstraint, SmartQuery<? exten
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/db/jdbc/OMA.java
```diff
@@ -81,6 +81,20 @@ public Database getDatabase(String realm) {
         return schema.getDatabase(realm);
     }
 
+    /**
+     * Provides the underlying database instance used to perform the actual statements.
+     * <p>
+     * Note that there is a helper available which can generate efficient UPDATE statements which are bound
+     * to various conditions. See {@link #updateStatement(Class)} for further information.
+     *
+     * @param entityType the entityto determine the database for
```
idlira, line 90:
```suggestion
     * @param entityType the entity to determine the database for
```

PR #306, opened 2020-04-10
Improvements for the QueryCompiler.

andyHa:
- Methods have been ordered to separate the
recursive descending parser from the constraint
generator.

- The lookup and generation of JOIN fetches for
complex property paths has been fixed.

- Temporal properties now support delta
expressions like "-2d" to filter the last
three days.


src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -13,27 +13,76 @@
...
+import java.time.temporal.TemporalUnit;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
- * Parses an SQL like query and compiles it into a {@link Constraint}.
+ * Parses a query and compiles it into a {@link Constraint}.
```
sabieber, line 39: This commit is really hard to review as it includes refactorings and logic changes in one. Also it breaks the build. Could this be revised?

src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -313,10 +319,21 @@ private boolean continueToken(boolean inQuotes) {
         return !reader.current().is(')', ':') && !reader.current().isWhitepace() && !isAtOperator();
     }
 
-    private Tuple<Mapping, Property> resolveProperty(String propertyPath) {
+    /**
+     * Resolves the given path into a property.
+     * <p>
+     * If the path contains ".", it will also yield an appropriate mapping which represents the necessarrry joins.
```
idlira, line 325:
```suggestion
     * If the path contains ".", it will also yield an appropriate mapping which represents the necessary joins.
```

src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -344,7 +361,28 @@ private boolean continueToken(boolean inQuotes) {
...
+     * <p>
+     * If the property cannot be resolved, the given <tt>token</tt> is passed into {@link #compileCustomField(String)}
+     * so that a sub class of the compiler can generate a constraint for a virtual field. If this also doesn't yield
+     * a constraint, a regular search in the default fields is generated. This might be necessarry so that tokens like
+     * an:value can be used as search term as long as no property named "an" exists.
+     *
+     * @param mappingAndProperty the mapping (join path) and the resolved property to address
+     * @param token              the token which has been parsed as field / property name
+     * @param skipped            <tt>true</tt> if we previously skilled whitespaces. Therefor fore "x=y" we generate
+     *                           this as literal term. But for "x =y" we simply generate a search for "x" and leave
```
idlira, line 379:
```suggestion
     *                           this as literal term. But for "x=y" we simply generate a search for "x" and leave
```

andyHa: Nope, this whitespace is put there on purpose... as "x =y" and "x=y" ist tokenized differently as long as x isn't a field / property.

PR #308, opened 2020-04-13
Replaces reflection calls with method handles.

andyHa:
With method handles, all access checks are performed when the
schema / EntityDescriptor is built. After that (when the handlers
are invoked a bazillion times) nothing will happen and the methods
are more or less directly invoked and also represented as a single
frame in a stack trace).

This should provide quite some performance improvements and
also increase the readability of stack traces.

### DIFF
--- a/src/main/java/sirius/db/mixing/EntityDescriptor.java
+++ b/src/main/java/sirius/db/mixing/EntityDescriptor.java
@@ -362,8 +362,10 @@ public class EntityDescriptor {
      * @param entity the entity to perform the handlers on
      */
     public final void beforeSave(Object entity) {
-        for (Consumer<Object> c : getSortedBeforeSaveHandlers()) {
-            c.accept(entity);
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/db/mixing/EntityDescriptor.java
```diff
@@ -592,21 +605,58 @@ private static void processMethod(EntityDescriptor descriptor, AccessPath access
...
+            } else if (method.getParameterCount() == 1) {
+                return entity -> {
+                    try {
+                        handle.invoke(accessPath.apply(entity), entity);
+                    } catch (Throwable e) {
+                        throw Exceptions.handle(Mixing.LOG, e);
+                    }
+                };
+            } else {
+                Mixing.LOG.WARN("Handler %s.%s must have either 0 or 1 Parrameters (the entity itself)!",
```
idlira, line 646:
```suggestion
                Mixing.LOG.WARN("Handler %s.%s must have either 0 or 1 Parameters (the entity itself)!",
```

PR #319, opened 2020-05-06
BREAKING CHANGE: Makes DateRange truly immutable.

andyHa:
DateRange looks and feels like a value class (which it
should be). Therefore the mutable flag was moved into
a method parameter, therefore no real functionality changed.

Also, some additional default ranges were added.

### DIFF
--- a/src/main/java/sirius/db/mixing/DateRange.java
+++ b/src/main/java/sirius/db/mixing/DateRange.java
@@ -272,17 +272,17 @@ public class DateRange {
      * @param useLocalDate determines if a {@link LocalDate} instead of a {@link LocalDateTime} should be applied.
      *                     This might be crucial as a <tt>LocalDateTime</tt> is encoded entirely differently
      *                     (see {@link sirius.db.jdbc.Databases#convertValue(Object))}.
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mixing/DateRange.java
```diff
@@ -211,36 +228,53 @@ public static DateRange beforeLastYear() {
...
-     * @param qry   the query to expand
+     * @param <C>          the constraint type to generate
+     * @param field        the field to filter on
+     * @param useLocalDate determines if a {@link LocalDate} instead of a {@link LocalDateTime} should be applied.
+     *                     This might be crucial as a <tt>LocalDateTime</tt> is encoded entirely differently
+     *                     (see {@link sirius.db.jdbc.Databases#convertValue(Object))}.
+     * @param qry          the query to expand
      */
-    public <C extends Constraint> void applyTo(String field, Query<?, ?, C> qry) {
+    public <C extends Constraint> void applyTo(String field, boolean useLocalDate, Query<?, ?, C> qry) {
```
bwiedmann, line 277: 🥇

PR #330, opened 2020-05-22
Permits to specify a callback for each delete.

andyHa:


src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -971,8 +973,13 @@ private long performScrollMonitoring(long lastScroll) {
...
 
     @Override
-    public void delete() {
-        iterateAll(elastic::delete);
+    public void delete(@Nullable Consumer<E> entityCallback) {
+        iterateAll(entity -> {
+            if (entityCallback != null) {
+                entityCallback.accept(entity);
+                elastic.delete(entity);
+            }
```
sabieber, line 981:
```suggestion
            if (entityCallback != null) {
                entityCallback.accept(entity);
            }
            elastic.delete(entity);
            
```

PR #339, opened 2020-06-17
Provides a common interface shared by all entities.

andyHa:
This simplifies defining database independent entities.
Examples would be UserAccount or CodeList in sirius-biz.

### DIFF
--- a/src/main/java/sirius/db/mixing/BaseEntity.java
+++ b/src/main/java/sirius/db/mixing/BaseEntity.java
@@ -41,7 +41,7 @@ import java.util.function.Consumer;
  *
  * @param <I> the type of the ID used by subclasses
  */
-public abstract class BaseEntity<I> extends Mixable {
+public abstract class BaseEntity<I> extends Mixable implements Entity {
 
     @Part
     protected static Mixing mixing;
@@ -87,29 +87,17 @@ public abstract class BaseEntity<I> extends Mixable {
      *
      * @return <tt>true</tt> if the entity has not been written to the database yes, <tt>false</tt> otherwise
      */
+    @Override
     public boolean isNew() {
         return getId() == null;
     }
 
-    /**
-     * Each entity type can be addressed by its class or by a unique name, which is its simple class name in upper
-     * case.
-     *
-     * @return the type name of this entity type
-     * @see #getUniqueName()
-     */
+    @Override
     public String getTypeName() {
         return Mixing.getNameForType(getClass());
     }
 
-    /**
-     * Returns an unique name of this entity.
-     * <p>
-     * This unique string representation of this entity is made up of its type along with its id.
-     *
-     * @return an unique representation of this entity or an empty string if the entity was not written to the
-     * database yet
-     */
+    @Override
     public final String getUniqueName() {
         if (isNew()) {
             return "";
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mixing/Entity.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package sirius.db.mixing;
+
+/**
+ * Describes methods implemented in {@link BaseEntity} and therefore available for all entities.
+ * <p>
+ * The reason to define an extra interface is to have a base interface when defining database independent entities.
+ * These are commonly defined as interface and then implemented by one or more subclasses of specific entities.
+ * However e.g. the templates often refert to the interface itself so there is only on UI to manage all kinds of
```
jmuscireum, line 16:
```suggestion
 * However e.g. the templates often referred to the interface itself so there is only one UI to manage all kinds of
```

src/main/java/sirius/db/mixing/Entity.java
```diff
@@ -0,0 +1,64 @@
...
+ * entities. Therefore we need a super interface so that tagliatelle (which only sees the interface) knows which
+ * methods are available.@
```
jmuscireum, line 18:
```suggestion
 * methods are available.
```

src/main/java/sirius/db/mixing/BaseEntity.java
```diff
@@ -87,29 +87,17 @@ public EntityDescriptor getDescriptor() {
      *
      * @return <tt>true</tt> if the entity has not been written to the database yes, <tt>false</tt> otherwise
```
jmuscireum, line 88:
```suggestion
     * @return <tt>true</tt> if the entity has not been written to the database yet, <tt>false</tt> otherwise
```

src/main/java/sirius/db/mixing/Entity.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public interface Entity {
+
+    /**
+     * Determines if the entity is new (not yet written to the database).
+     *
+     * @return <tt>true</tt> if the entity has not been written to the database yes, <tt>false</tt> otherwise
```
jmuscireum, line 25:
```suggestion
     * @return <tt>true</tt> if the entity has not been written to the database yet, <tt>false</tt> otherwise
```

src/main/java/sirius/db/mixing/Entity.java
```diff
@@ -0,0 +1,64 @@
...
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.db.mixing;
+
+/**
+ * Describes methods implemented in {@link BaseEntity} and therefore available for all entities.
+ * <p>
+ * The reason to define an extra interface is to have a base interface when defining database independent entities.
+ * These are commonly defined as interface and then implemented by one or more subclasses of specific entities.
```
mkeckmkeck, line 15:
```suggestion
 * They are commonly defined as interfaces and then implemented by one or more subclasses of specific entities.
```

src/main/java/sirius/db/mixing/Entity.java
mkeckmkeck, line 16:
```suggestion
 * However e.g. the templates often refer to the interface itself so there is only one UI to manage all kinds of
```

src/main/java/sirius/db/mixing/Entity.java
```diff
@@ -0,0 +1,64 @@
...
+ * However e.g. the templates often refert to the interface itself so there is only on UI to manage all kinds of
+ * entities. Therefore we need a super interface so that tagliatelle (which only sees the interface) knows which
```
mkeckmkeck, line 17:
```suggestion
 * entities. Therefore we need a super interface so that Tagliatelle (which only sees the interface) knows which
```

src/main/java/sirius/db/mixing/Entity.java
mkeckmkeck, line 18:
```suggestion
 * methods are available.
```

PR #347, opened 2020-07-03
Adds elastic query helpers for adding an aggregation counting total hits

sabieber:
This can be necessary as the normal total hits field returned bei Elasticsearch will only count up to 10.000. For scenarios where more entities that match a query have to be counted the new methods can nbe used.

Fixes: OX-5984

### DIFF
--- a/src/main/java/sirius/db/es/ElasticQuery.java
+++ b/src/main/java/sirius/db/es/ElasticQuery.java
@@ -62,6 +62,7 @@ public class ElasticQuery<E extends ElasticEntity> extends Query<ElasticQuery<E>
     private static final String KEY_FIELD = "field";
     private static final String KEY_TERMS = "terms";
     private static final String KEY_CARDINALITY = "cardinality";
+    private static final String KEY_VALUE_COUNT = "value_count";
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -942,14 +954,39 @@ public JSONObject getRawResponse() {
     }
 
     /**
-     * Returns the total number of hits for this query.
+     * Returns the total number of hits (with a maximum of 10000) for this query.
+     * <p>
+     * If more than 10000 hits are expected an appropriate aggregation should be {@link #addTotalHitsAggregation() added}
+     * and {@link #getAggregatedTotalHits() evaluated} instead to get the real count.
```
idlira, line 960:
```suggestion
     * and {@link #getAggregatedTotalHits() evaluated} instead of getting the real count.
```

sabieber, line 960:
```suggestion
     * and {@link #getAggregatedTotalHits() evaluated} instead, to get the real count.
```

PR #357, opened 2020-07-15
Fixes an NPE and adds a boilerplate getter.

andyHa:


src/main/java/sirius/db/es/AggregationBuilder.java
```diff
@@ -368,7 +372,7 @@ public String getName() {
      * @return <tt>true</tt> if there are sub aggregations, <tt>false</tt> otherwise
      */
     public boolean hasSubAggregations() {
-        return !subAggregations.isEmpty();
+        return subAggregations != null && !subAggregations.isEmpty();
```
qw3ry, line 375:
```suggestion
        return !getSubAggregations().isEmpty();
```

PR #362, opened 2020-07-23
BREAKING: Changes FieldLookupCache so it always returns the transformed value

meg1502:
Previously, the raw database value was cached and returned, but now we use the property to transform the value into the actual field type


src/main/java/sirius/db/mixing/FieldLookupCache.java
```diff
@@ -47,16 +46,16 @@
      * @param id    the id of the entity to resolve
      * @param field the field to resolve
      * @param <E>   the generic type of the entitiy
-     * @return the value of the field or an empty value if either the field is empty or the given ID was <tt>null</tt>
+     * @return the value of the field or null if either the field is empty or the given ID was <tt>null</tt>
      */
-    public <E extends BaseEntity<?>> Value lookup(Class<E> type, Object id, Mapping field) {
+    public <E extends BaseEntity<?>> Object lookup(Class<E> type, Object id, Mapping field) {
```
andyHa, line 51: why change the outside API and not simply wrap the result in a value to provide a non-null method?

src/test/java/sirius/db/mixing/FieldLookupCacheSpec.groovy
```diff
@@ -0,0 +1,74 @@
...
+import sirius.db.jdbc.OMA
+import sirius.db.jdbc.SmartQueryTestEntity
+import sirius.db.mongo.Mango
+import sirius.db.mongo.MangoTestEntity
+import sirius.kernel.BaseSpecification
+import sirius.kernel.di.std.Part
+
+import java.time.LocalDateTime
+
+class FieldLookupCacheSpec extends BaseSpecification {
```
andyHa, line 20: maybe add a test with:
- a field in a composite
- a field in a mixin
- a composite field in a mixin

these are common problems when working with transformFromDatabase etc. as it is quite easy to get the accessPath wrong....

src/main/java/sirius/db/mixing/BaseMapper.java
```diff
@@ -550,7 +550,7 @@ public B resolveOrFail(String name, ContextInfo... info) {
      * @param type  the type of the entity
      * @param id    the id of the entity
      * @param field the field to resolve
-     * @return the field value
+     * @return the field value, transformed into the appropiate type
```
bwiedmann, line 553:
```suggestion
     * @return the field value, transformed into the appropriate type
```

src/main/java/sirius/db/mixing/FieldLookupCache.java
```diff
@@ -51,7 +51,7 @@
      */
     public <E extends BaseEntity<?>> Value lookup(Class<E> type, Object id, Mapping field) {
         if (Strings.isEmpty(id)) {
-            return Value.EMPTY;
+            return null;
```
sabieber, line 54:
```suggestion
            return Value.EMPTY;
```

src/main/java/sirius/db/mixing/FieldLookupCache.java
```diff
@@ -73,7 +73,7 @@
                               id,
                               type)
                       .handle();
-            return Value.EMPTY;
+            return null;
```
sabieber, line 76:
```suggestion
            return Value.EMPTY;
```

PR #364, opened 2020-07-26
Fixes for long running Clickhouse Queries...

andyHa:
- Matches the client socket timeout with the server timeout (300s instead of 30s)
- Permits to suppress warnings for queries which are expected to run for a "long" time.


src/main/java/sirius/db/jdbc/WrappedPreparedStatement.java
```diff
@@ -42,24 +42,32 @@
...
+    private static final Duration QUERY_OPERATION = Duration.ofSeconds(30);
+
     private PreparedStatement delegate;
     private final String preparedSQL;
+    private boolean longRunning;
 
-    WrappedPreparedStatement(PreparedStatement prepareStatement, String preparedSQL) {
-        delegate = prepareStatement;
+    WrappedPreparedStatement(PreparedStatement prepareStatement, boolean longRunning, String preparedSQL) {
+        this.delegate = prepareStatement;
```
meg1502, line 54: prepare*d*Statement

src/main/java/sirius/db/jdbc/WrappedPreparedStatement.java
```diff
@@ -42,24 +42,32 @@
  * Wrapper for {@link PreparedStatement} to add microtiming.
  */
 class WrappedPreparedStatement implements PreparedStatement {
+
+    private static final Duration LONG_RUNNING_QUERY_OPERATION = Duration.ofMinutes(15);
```
meg1502, line 46: Shouldn't this be 5 minutes if trying to match 300s? 🧐

andyHa: No the actual socket timeout is specified in the config. However, as this is a hardcoded constant (for now), we permit an even longer interval...

PR #379, opened 2020-10-27
Prevents NPE when trying to read a MultiLanguageString with a null value

bwiedmann:
… now. This should not happen as null values should not be persisted and therefore not mapped when reading from database.

Fixes: SIRI-272
error

### DIFF
--- a/src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
+++ b/src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
@@ -111,7 +111,9 @@ public class MultiLanguageStringProperty extends BaseMapProperty implements ESPr
         Map<String, String> texts = new LinkedHashMap<>();
         for (Document document : (List<Document>) object.get()) {
             Object textValue = document.get(TEXT_PROPERTY);
-            texts.put(document.get(LANGUAGE_PROPERTY).toString(), textValue != null ? textValue.toString() : null);
+            if (textValue != null) {
+               texts.put(document.get(LANGUAGE_PROPERTY).toString(), ttextValue.toString());
+            }
         }
         return texts;
     }
--- a/src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
+++ b/src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
@@ -110,7 +110,8 @@ public class MultiLanguageStringProperty extends BaseMapProperty implements ESPr
     protected Object transformFromMongo(Value object) {
         Map<String, String> texts = new LinkedHashMap<>();
         for (Document document : (List<Document>) object.get()) {
-            texts.put(document.get(LANGUAGE_PROPERTY).toString(), document.get(TEXT_PROPERTY).toString());
+            Object textValue = document.get(TEXT_PROPERTY);
+            texts.put(document.get(LANGUAGE_PROPERTY).toString(), textValue != null ? textValue.toString() : null);
         }
         return texts;
     }
--- a/src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
+++ b/src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
@@ -67,8 +67,8 @@ public class MultiLanguageStringProperty extends BaseMapProperty implements ESPr
                            Consumer<Property> propertyConsumer) {
             if (!Modifier.isFinal(field.getModifiers())) {
                 Mixing.LOG.WARN("Field %s in %s is not final! This will probably result in errors.",
-                        field.getName(),
-                        field.getDeclaringClass().getName());
+                                field.getName(),
+                                field.getDeclaringClass().getName());
             }
 
             propertyConsumer.accept(new MultiLanguageStringProperty(descriptor, accessPath, field));
@@ -87,11 +87,11 @@ public class MultiLanguageStringProperty extends BaseMapProperty implements ESPr
             }
             if (supportedLanguages != null && !supportedLanguages.contains(language)) {
                 throw Exceptions.createHandled()
-                        .withNLSKey("MultiLanguageString.invalidLanguage")
-                        .set("language", language)
-                        .set("text", text)
-                        .set("field", getField().getName())
-                        .handle();
+                                .withNLSKey("MultiLanguageString.invalidLanguage")
+                                .set("language", language)
+                                .set("text", text)
+                                .set("field", getField().getName())
+                                .handle();
             }
         });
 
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
```diff
@@ -110,7 +110,8 @@ protected MultiLanguageString getMultiLanguageString(Object entity) {
     protected Object transformFromMongo(Value object) {
         Map<String, String> texts = new LinkedHashMap<>();
         for (Document document : (List<Document>) object.get()) {
-            texts.put(document.get(LANGUAGE_PROPERTY).toString(), document.get(TEXT_PROPERTY).toString());
+            Object textValue = document.get(TEXT_PROPERTY);
+            texts.put(document.get(LANGUAGE_PROPERTY).toString(), textValue != null ? textValue.toString() : null);
```
andyHa, line 114:
```suggestion
            if (textValue != null) {
               texts.put(document.get(LANGUAGE_PROPERTY).toString(), ttextValue.toString());
            }
```

src/main/java/sirius/db/mixing/properties/MultiLanguageStringProperty.java
```diff
@@ -110,7 +110,8 @@ protected MultiLanguageString getMultiLanguageString(Object entity) {
     protected Object transformFromMongo(Value object) {
         Map<String, String> texts = new LinkedHashMap<>();
         for (Document document : (List<Document>) object.get()) {
-            texts.put(document.get(LANGUAGE_PROPERTY).toString(), document.get(TEXT_PROPERTY).toString());
```
andyHa, line 113: note that this is not the original error - as the true error is storing a null value at all - still making this code more robust is a good idea.

andyHa (changes requested):
...

sabieber (approved):

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -77,6 +77,8 @@ public MultiLanguageString addText(String text) {
     public MultiLanguageString addText(String language, String text) {
         if (text != null) {
```
andyHa, line 78: duplicate check

src/main/java/sirius/db/mixing/types/SafeMap.java
```diff
@@ -55,13 +55,19 @@
...
+     * <br>
+     * Keys with null values will be removed.
      *
      * @param key   the key used to store the value
      * @param value the value to store
      * @return the map itself for fluent method calls
      */
     public SafeMap<K, V> put(@Nonnull K key, V value) {
-        modify().put(key, value);
+        if (value != null) {
```
andyHa, line 66: shouldn't we only overwrite this in MultiLanguageString to avoid unwanted side effects?

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -67,7 +67,7 @@ public MultiLanguageString addText(String text) {
     /**
      * Adds a new text for the given language.
      * <p>
-     * Null texts will be ignored.
+     * Texts with null values will be removed.
```
meg1502, line 70: there are other methods that call this method which have the same javadoc which needs to be adjusted

andyHa (commented):
..

src/main/java/sirius/db/mixing/types/SafeMap.java
```diff
@@ -55,13 +55,19 @@
...
+            modify().put(key, value);
+        } else {
+            modify().remove(key);
+        }
         return this;
```
meg1502, line 71: shouldn't we rather override this in MultiLanguageString than changing this in SafeMap?

meg1502 (changes requested):

bwiedmann:
Changing it directly in SafeMap was a dangerous mission - so overriding in MultiLanguageString should be safe and not changing the default behavior. Thanks!

andyHa (approved):

andyHa (approved):

sabieber (approved):

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -204,4 +204,32 @@ public String fetchTextOrFallback(String language) {
     private boolean hasFallback() {
         return withFallback && containsKey(FALLBACK_KEY);
     }
+
+    /**
+     * Puts the given key and value into the map.
+     * <br>
+     * Keys with null values will be removed.
```
meg1502, line 211:
```suggestion
     * If a null text is given it will be ignored, if the list already contains an entry it will be removed.
```

src/test/java/sirius/db/mongo/properties/MongoMultiLanguageStringPropertySpec.groovy
```diff
@@ -177,4 +177,60 @@ class MongoMultiLanguageStringPropertySpec extends BaseSpecification {
...
+    def "keys with null values are removed from the underlying map if a key already exists"() {
+        given:
+        CallContext.getCurrent().setLang("en")
+        def entity = new MongoMultiLanguageStringEntity()
+        entity.getMultiLangText().addText("en", "Super")
+        mango.update(entity)
+
+        when:
+        entity = mango.refreshOrFail(entity)
+        entity.getMultiLangText().fetchText() == "Super"
```
meg1502, line 209: it's a bit confusing to have a condition in a 'when' block and i'm not sure if this will result in a test failure because its not in a then block (or and: chained to a then block)

meg1502 (changes requested):
It's still possible to add null values by obtaining the underlying map with .modify - should we block its access using an UnsupportedOperationException?

bwiedmann:
> It's still possible to add null values by obtaining the underlying map with .modify - should we block its access using an UnsupportedOperationException?

Good hint. What about removing keys from the underlying list as done eg. in `sirius.db.es.properties.ESStringLocalDateTimePropertySpec#reading and writing works for Elasticsearch`? Should we introduce a `sirius.db.mixing.types.MultiLanguageString#remove` method calling `super.modify().remove()` here`

src/test/java/sirius/db/mongo/properties/MongoMultiLanguageStringPropertySpec.groovy
bwiedmann: Thx! Fixed now.

src/main/java/sirius/db/mixing/types/MultiLanguageString.java
```diff
@@ -232,4 +232,28 @@ public void setData(Map<String, String> newData) {
                              .filter(entry -> entry.getValue() != null)
                              .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
     }
+
+    @Override
+    public Map<String, String> modify() {
```
meg1502, line 237: javadoc :)

meg1502 (changes requested):

PR #380, opened 2020-10-29
Falls back to estimated document count for queries with no filters

meg1502:
Also offers some new performance related options for count queries

### DIFF
--- a/src/main/java/sirius/db/mongo/Finder.java
+++ b/src/main/java/sirius/db/mongo/Finder.java
@@ -389,14 +389,14 @@ public class Finder extends QueryBuilder<Finder> {
      * If the there are no filters in this query and forceAccurate is false, a pre-counted estimate is returned instead.
      *
      * @param collection    the collection to search in
-     * @param forceAccurate always count the actual query using countDocuments
+     * @param forceAccurate if set to <tt>true</tt> we'll never use <b>estimatedDocumentCount</b> which is way more efficient but might return wrong values in case a cluster is active which had experienced an unclean shutdown.
      * @param maxTimeMS     the maximum process time for this cursor in milliseconds, 0 for unlimited
      * @return the number of documents found
      */
     public long countIn(String collection, boolean forceAccurate, long maxTimeMS) {
         Watch w = Watch.start();
         try {
-            if (filterObject.size() == 0 && !forceAccurate) {
+            if (filterObject.isEmpty() && !forceAccurate) {
                 return mongo.db().getCollection(collection).estimatedDocumentCount(new EstimatedDocumentCountOptions().maxTime(maxTimeMS, TimeUnit.MILLISECONDS));
             }
             return mongo.db()
--- a/src/main/java/sirius/db/mongo/Finder.java
+++ b/src/main/java/sirius/db/mongo/Finder.java
@@ -9,6 +9,7 @@
 package sirius.db.mongo;
 
 import com.mongodb.BasicDBObject;
+import com.mongodb.MongoExecutionTimeoutException;
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoCursor;
 import com.mongodb.client.MongoIterable;
@@ -22,6 +23,7 @@ import sirius.kernel.async.TaskContext;
 import sirius.kernel.commons.Monoflop;
 import sirius.kernel.commons.Value;
 import sirius.kernel.commons.Watch;
+import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.Microtiming;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -371,16 +373,35 @@ public long countIn(Class<?> type) {
...
+     *
+     * @param collection    the collection to search in
+     * @param forceAccurate always count the actual query using countDocuments
+     * @param maxTimeMS     the maximum process time for this cursor in milliseconds, 0 for unlimited
+     * @return the number of documents found
+     */
+    public long countIn(String collection, boolean forceAccurate, long maxTimeMS) {
         Watch w = Watch.start();
         try {
+            if (filterObject.size() == 0 && !forceAccurate) {
```
andyHa, line 399:
```suggestion
            if (filterObject.isEmpty() && !forceAccurate) {
```

src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -371,16 +373,35 @@ public long countIn(Class<?> type) {
...
      */
     public long countIn(String collection) {
+        return countIn(collection, false, 0);
+    }
+
+    /**
+     * Counts the number of documents in the result of the given query.
+     * <p>
+     * Note that limits are ignored for this query.
+     * If the there are no filters in this query and forceAccurate is false, a pre-counted estimate is returned instead.
```
mkeckmkeck, line 389:
```suggestion
     * If there are no filters in this query and forceAccurate is false, a pre-counted estimate is returned instead.
```

src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -371,16 +373,35 @@ public long countIn(Class<?> type) {
...
+     *
+     * @param collection    the collection to search in
+     * @param forceAccurate always count the actual query using countDocuments
```
andyHa, line 392:
```suggestion
     * @param forceAccurate if set to &lt;tt&gt;true&lt;/tt&gt; we'll never use &lt;b&gt;estimatedDocumentCount&lt;b&gt; which is way more efficient but might return wrong values in case a cluster is active which had experienced an unclean shutdown.
```

andyHa (approved):
Approved - only minor javadoc hints :)

mkeckmkeck (approved):

src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -371,16 +375,43 @@ public long countIn(Class<?> type) {
...
         Watch w = Watch.start();
         try {
-            return mongo.db()
-                        .getCollection(collection)
-                        .countDocuments(filterObject, new CountOptions().collation(mongo.determineCollation()));
+            if (filterObject.isEmpty() && !forceAccurate) {
+                return Optional.of(mongo.db()
+                                        .getCollection(collection)
+                                        .estimatedDocumentCount(new EstimatedDocumentCountOptions().maxTime(maxTimeMS,
+                                                                                                            TimeUnit.MILLISECONDS)));
```
meg1502, line 405: the max time for the estimate is not really useful, as even with a huge collection this is always super fast and there is probably not even an interrupt point in the operation, as even with 1ms this never hit the time out - but i think it's still a good idea to include as the option is available

src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -379,29 +381,37 @@ public long countIn(Class<?> type) {
...
-     * @return the number of documents found
+     * @return the number of documents found, wrapped in an Optional, or an empty Optional if the query timed out
      */
-    public long countIn(String collection, boolean forceAccurate, long maxTimeMS) {
+    public Optional<Long> countIn(String collection, boolean forceAccurate, long maxTimeMS) {
         Watch w = Watch.start();
         try {
             if (filterObject.isEmpty() && !forceAccurate) {
-                return mongo.db().getCollection(collection).estimatedDocumentCount(new EstimatedDocumentCountOptions().maxTime(maxTimeMS, TimeUnit.MILLISECONDS));
+                return Optional.of(mongo.db()
```
sabieber, line 402: Can the Mongo Methods return null? In that case it should be `ofNullable` instead of `of` as it will throw an NPE otherwise.

sabieber (changes requested):

src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -371,16 +375,43 @@ public long countIn(Class<?> type) {
      * Counts the number of documents in the result of the given query.
      * <p>
      * Note that limits are ignored for this query.
+     * If the there are no filters in this query, an estimate is returned instead.
```
idlira, line 378:
```suggestion
     * If there are no filters in this query, an estimation is returned instead.
```

src/main/java/sirius/db/mongo/Finder.java
meg1502: these methods return a primitive long

src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -371,16 +375,43 @@ public long countIn(Class<?> type) {
...
      */
     public long countIn(String collection) {
+        return countIn(collection, false, 0).orElse(0L);
+    }
+
+    /**
+     * Counts the number of documents in the result of the given query.
+     * <p>
+     * Note that limits are ignored for this query.
+     * If there are no filters in this query and forceAccurate is false, a pre-counted estimate is returned instead.
```
idlira, line 391:
```suggestion
     * If there are no filters in this query and forceAccurate is false, a pre-counted estimation is returned instead.
```

idlira (approved):
Small typos

src/main/java/sirius/db/mongo/Finder.java
meg1502: estimate is correct here.

src/main/java/sirius/db/mongo/Finder.java
meg1502: estimate is correct here

bwiedmann (approved):

PR #381, opened 2020-11-10
Fixes DateRange by using suppliers as they are often used in a static context

fhaScireum:
- FIxes: OX-6320

### DIFF
--- a/src/main/java/sirius/db/mixing/DateRange.java
+++ b/src/main/java/sirius/db/mixing/DateRange.java
@@ -30,25 +30,25 @@ import java.util.function.Supplier;
 public class DateRange {
 
     private final String key;
-    private final String name;
+    private final Supplier<String> nameSupplier;
     private final Supplier<LocalDateTime> fromSupplier;
     private final Supplier<LocalDateTime> untilSupplier;
 
     /**
-     * Creates a new DateRange with the given unique key, translated (shown) name and two date suppliers
-     * specifying the range
+     * Creates a new DateRange with the given unique key, a supplier that returns the translated (shown) name
+     * and two date suppliers specifying the range
      *
      * @param key           the unique name of the ranged used as filter value
-     * @param name          the translated name shown to the user
+     * @param nameSupplier          the translated name shown to the user
      * @param fromSupplier  the lower limit (including) of the range
      * @param untilSupplier the upper limit (including) of the range
      */
     public DateRange(String key,
-                     String name,
+                     Supplier<String> nameSupplier,
                      @Nullable Supplier<LocalDateTime> fromSupplier,
                      @Nullable Supplier<LocalDateTime> untilSupplier) {
         this.key = key;
-        this.name = name;
+        this.nameSupplier = nameSupplier;
         this.fromSupplier = fromSupplier;
         this.untilSupplier = untilSupplier;
     }
@@ -60,7 +60,7 @@ public class DateRange {
      */
     public static DateRange lastFiveMinutes() {
         return new DateRange("5m",
-                             NLS.get("DateRange.5m"),
+                             () -> NLS.get("DateRange.5m"),
                              () -> LocalDateTime.now().minusMinutes(5),
                              LocalDateTime::now);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/mixing/DateRange.java
```diff
@@ -276,17 +294,19 @@ public LocalDateTime getUntil() {
      */
     public <C extends Constraint> void applyTo(String field, boolean useLocalDate, Query<?, ?, C> query) {
         List<C> constraints = new ArrayList<>(2);
-        if (from != null) {
+        if (fromSupplier != null) {
```
sabieber, line 297: hier sollten auch die getFrom und getUntil aufgerufen werden und im if geprüft werden ob die resultierenden localdatetimes null sind. sonst ändert sich das verhalten durch den PR

sabieber (changes requested):

src/main/java/sirius/db/mixing/DateRange.java
```diff
@@ -248,7 +266,7 @@ public String getKey() {
      * @return the start of the date range
      */
     public LocalDateTime getFrom() {
-        return from;
+        return fromSupplier != null ? fromSupplier.get() : null;
```
idlira, line 269: What situation could actually bring such a supplier to be null?

src/main/java/sirius/db/mixing/DateRange.java
idlira: What could make such a supplier return null in first place?

idlira (changes requested):
...

src/main/java/sirius/db/mixing/DateRange.java
fhaScireum: If the constructor would be called without a from supplier :)

fhaScireum:
> A small improvement to the class could be a `.withReferenceDate` which is used if defined instead of LocalDateTime.now(), allowing a caller to actually freeze dates if really desired.

In that case the user could call the constructor with the until or from supplier always returning a specific date. I think calling a quasi constructors with a descriptive method name and overruling one of the two essential dates is quite messy.

idlira (approved):

src/main/java/sirius/db/mixing/DateRange.java
sabieber: Dont know, but better safe than sorry. The null checks are already in place

sabieber (approved):

src/main/java/sirius/db/mixing/DateRange.java
```diff
@@ -55,7 +59,10 @@ public DateRange(String key, String name, @Nullable LocalDateTime from, @Nullabl
      * @return a date range for the given interval
      */
     public static DateRange lastFiveMinutes() {
```
andyHa, line 61: shouldn't this be constants then? (all factory methods)

andyHa: maybe we should add constants and make the methods deprecated as a first step

src/main/java/sirius/db/mixing/DateRange.java
```diff
@@ -29,24 +30,27 @@
...
-     * @param key   the unique name of the ranged used as filter value
-     * @param name  the translated name shown to the user
-     * @param from  the lower limit (including) of the range
-     * @param until the upper limit (including) of the range
+     * @param key           the unique name of the ranged used as filter value
+     * @param nameSupplier  the translated name shown to the user
+     * @param fromSupplier  the lower limit (including) of the range
+     * @param untilSupplier the upper limit (including) of the range
      */
-    public DateRange(String key, String name, @Nullable LocalDateTime from, @Nullable LocalDateTime until) {
```
andyHa, line 45: add a legacy constructor which invokes a new one and mark this deprecated? this way we prevent a breaking change...

andyHa (changes requested):
...

src/main/java/sirius/db/mixing/DateRange.java
```diff
@@ -29,138 +30,206 @@
...
 
     private final String key;
-    private final String name;
-    private final LocalDateTime from;
-    private final LocalDateTime until;
+    private final Supplier<String> nameSupplier;
+    private final Supplier<LocalDateTime> fromSupplier;
+    private final Supplier<LocalDateTime> untilSupplier;
+
+    public static final DateRange LAST_FIVE_MINUTES = lastFiveMinutes();
```
andyHa, line 37: javadoc + don't call depracated code here...

idlira (approved):

andyHa (changes requested):
...

PR #382, opened 2020-11-10
Permits to read from secondary nodes in MongoDB

andyHa:
When using findInSecondary or selectFromSecondary we now issue a ReadPreference "NEAREST" instead of
"PRIMARY".

### DIFF
--- a/src/main/java/sirius/db/mongo/Mango.java
+++ b/src/main/java/sirius/db/mongo/Mango.java
@@ -234,7 +234,7 @@ public class Mango extends BaseMapper<MongoEntity, MongoConstraint, MongoQuery<?
      * This provides an essential boost in performance, as all nodes of a MongoDB cluster are utilized. However, this
      * may return stale data if a secondary lags behind. Therefore this data must not be stored back in the primary
      * database using {@link Mango#update(MongoEntity)}. This should rather only be used to serve web requests or other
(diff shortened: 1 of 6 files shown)


sabieber (approved):

src/main/java/sirius/db/mongo/Mango.java
```diff
@@ -225,7 +225,25 @@ protected void deleteEntity(MongoEntity entity, boolean force, EntityDescriptor
...
+        return new MongoQuery<>(mixing.getDescriptor(type), null);
+    }
+
+    /**
+     * In contrast to {@link #select(Class)} this doesn't necessarily read from the primary node but from the nearest.
+     * <p>
+     * This provides an essential boost in performance, as all nodes of a MongoDB cluster are utilized. However, this
+     * may return stale data if a secondary lags behind. Therefore this data must not be stored back in the primary
+     * database using {@link Mango#update(MongoEntity)}. This should rather only be used to serve web requests or other
+     * queries where occasional stale date does no harm.
```
idlira, line 237:
```suggestion
     * queries where occasional stale data does no harm.
```

idlira (changes requested):
Small typo.
Shouldn't we expose this feature in mango.select as well?

mkeckmkeck (approved):

PR #385, opened 2020-11-16
No Normalisation in Search Index

jakobvogel:
Removes character reduction stage from tokenizer, due to an incompatibility between search index and query normalisation. (See code.)

### DIFF
--- a/src/main/java/sirius/db/text/BasicIndexTokenizer.java
+++ b/src/main/java/sirius/db/text/BasicIndexTokenizer.java
@@ -18,10 +18,11 @@ public class BasicIndexTokenizer extends Tokenizer {
     @Override
     protected ChainableTokenProcessor createProcessor() {
         // ATTENTION: The pipeline used to contain a "new ReduceCharacterProcessor()" stage. This lead to the search
-        // index containing normalised strings, unlike earlier verions of sirius-db. Consequently, MEMOIO was no longer
-        // able to find search terms containing umlauts. Weighing the two possible options — keeping the normalised
-        // index and normalising the search term, or reverting the index – the decision was made to remove search index
-        // normalisation.
+        // index consisting of normalised strings, unlike earlier verions of sirius-db. Consequently, without migration,
+        // existing indices were incompatible. Furthermore, ElasticQueryCompiler did not follow along and continued to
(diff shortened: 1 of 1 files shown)


idlira (changes requested):
Just not sure if we should place such comments in sirius code, as we mention own products here...

jakobvogel:
> Just not sure if we should place such comments in sirius code, as we mention own products here...

@andyHa

src/main/java/sirius/db/text/BasicIndexTokenizer.java
```diff
@@ -17,8 +17,14 @@
 
     @Override
     protected ChainableTokenProcessor createProcessor() {
+        // ATTENTION: The pipeline used to contain a "new ReduceCharacterProcessor()" stage. This lead to the search
+        // index containing normalised strings, unlike earlier verions of sirius-db. Consequently, MEMOIO was no longer
+        // able to find search terms containing umlauts. Weighing the two possible options — keeping the normalised
+        // index and normalising the search term, or reverting the index – the decision was made to remove search index
+        // normalisation.
+        // --- Stage removed by JVO on 2020/11/16 after discussion with AHA
```
andyHa, line 25: this line can be removed - this is stored in git already :)

src/main/java/sirius/db/text/BasicIndexTokenizer.java
```diff
@@ -17,8 +17,14 @@
 
     @Override
     protected ChainableTokenProcessor createProcessor() {
+        // ATTENTION: The pipeline used to contain a "new ReduceCharacterProcessor()" stage. This lead to the search
+        // index containing normalised strings, unlike earlier verions of sirius-db. Consequently, MEMOIO was no longer
```
andyHa, line 21: this wasn't just a MEMOIO problem - essentially the ElasticQueryCompiler doesn't perform the same reduction - therefore all terms with appropriate characters became problematic / unfindable / whatever :-)

jakobvogel: revised

andyHa (approved):

idlira (approved):

qw3ry (approved):

PR #389, opened 2020-12-10
Fix: Query Compilation when using References (SIRI-288)

jakobvogel:
> https://scireum.myjetbrains.com/youtrack/issue/SIRI-288

- Protects query compilation against exceptions

### DIFF
--- a/src/main/java/sirius/db/mixing/query/QueryCompiler.java
+++ b/src/main/java/sirius/db/mixing/query/QueryCompiler.java
@@ -24,6 +24,7 @@ import sirius.kernel.commons.Tuple;
 import sirius.kernel.commons.Value;
 import sirius.kernel.di.GlobalContext;
 import sirius.kernel.di.std.Part;
+import sirius.kernel.health.Exceptions;
 
 import javax.annotation.Nullable;
 import java.io.StringReader;
@@ -724,6 +725,7 @@ public abstract class QueryCompiler<C extends Constraint> {
             // second line of defence: should the transformation fail for whatever reason, we pass the original string
             // as is, leaving it to the database to come up with no results if searching for an invalid value (such as a
             // string in a numeric field)
+            Exceptions.ignore(e);
             return Value.of(value);
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -710,7 +711,21 @@ protected Object compileStringValue(Property property, String value) {
...
+        }
+
+        try {
+            return property.transformValue(Value.of(value));
+        } catch (Exception e) {
+            // second line of defence: should the transformation fail for whatever reason, we pass the original string
+            // as is, leaving it to the database to come up with no results if searching for an invalid value (such as a
+            // string in a numeric field)
+            return Value.of(value);
+        }
```
jakobvogel, line 728: Do we want to do this? Originally, when looking for a string in an integer field (for example), there was an error that the type is illegal. With the try/catch, we leave it to the database to handle that case.

andyHa: I'd keep this in place but also add "Exceptions.ignore(e)" this signals that the ex is deliberately ignore and also - for the desperate - there is a logger to enable und view these exception

jakobvogel: Done 👍

andyHa (approved):

src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -710,7 +712,22 @@ protected Object compileStringValue(Property property, String value) {
             }
         }
 
-        return property.transformValue(Value.of(value));
+        // first line of defence: reference types are not processed as transformation triggers a lookup in the
```
fhaScireum, line 715:
```suggestion
        // first line of defence: reference types are not processed as transformation triggers at lookup in the
```

fhaScireum (approved):

src/main/java/sirius/db/mixing/query/QueryCompiler.java
jakobvogel: Nope. 🙂 The transformation _here_ (from `String` to `Value`) triggers the lookup in the _other_ database (where the reference in question points to), not the other way around.

sabieber (approved):

cschierle (approved):

mkeckmkeck (approved):

PR #392, opened 2020-12-21
Increases the timeouts and fetch sizes for scroll queries.

andyHa:
Only fetching 10 or 50 entities results in a very poor performance
as the duration of the scroll fetch queries dominate the runtime.

We thus greatly increased the timeout (which shouldn't matter at
all, as we (under normal conditions) close a scroll once completed).

Within the now longer timeout, we can process way more entities
within one "batch".

### DIFF
--- a/src/main/java/sirius/db/es/ElasticQuery.java
+++ b/src/main/java/sirius/db/es/ElasticQuery.java
@@ -58,9 +58,24 @@ public class ElasticQuery<E extends ElasticEntity> extends Query<ElasticQuery<E>
     @Deprecated
     public static final int DEFAULT_TERM_AGGREGATION_BUCKET_COUNT = 25;
 
-    private static final int SCROLL_TTL_SECONDS = 60 * 5;
-    private static final int MAX_SCROLL_RESULTS_FOR_SINGLE_SHARD = 50;
-    private static final int MAX_SCROLL_RESULTS_PER_SHARD = 10;
+    /**
+     * This is the timeout we specify for elastic search for a scroll requests.
+     * <p>
+     * This essentially instructs ES to keep a scroll response open for the given timeout
+     */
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -58,9 +58,24 @@
...
-    private static final int MAX_SCROLL_RESULTS_PER_SHARD = 10;
+    /**
+     * This is the timeout we specify for elastic search for a scroll requests.
+     * <p>
+     * This essentially instructs ES to keep a scroll response open for the given timeout
+     */
+    private static final int SCROLL_TTL_SECONDS = 60 * 15;
+
+    /**
+     * If we only fetch from a single shard (as we use a routing), we fetch up to 800 entities at once and hope to
```
qw3ry, line 69:
```suggestion
     * If we only fetch from a single shard (as we use a routing), we fetch up to {@link #MAX_SCROLL_RESULTS_FOR_SINGLE_SHARD} entities at once and hope to
```

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -58,9 +58,24 @@
...
+     * process them within {@link #SCROLL_TTL_SECONDS}.
+     */
+    private static final int MAX_SCROLL_RESULTS_FOR_SINGLE_SHARD = 800;
+
+    /**
+     * If we fetch from many shards, we fetch up to 100 entities per shards and hope to process them within
```
qw3ry, line 75:
```suggestion
     * If we fetch from many shards, we fetch up to {@link #MAX_SCROLL_RESULTS_PER_SHARD} entities per shards and hope to process them within
```

qw3ry (approved):

PR #416, opened 2021-03-18
Extends SkipDefaultValue to empty maps and lists

idlira:
So we can annotate map and list based fields to not be physically persisted when their contents are empty.

Direct use case are MultiLanguageStrings.

### DIFF
--- a/src/main/java/sirius/db/mongo/Mango.java
+++ b/src/main/java/sirius/db/mongo/Mango.java
@@ -33,6 +33,8 @@ import sirius.kernel.health.Exceptions;
 
 import java.util.HashSet;
 import java.util.IntSummaryStatistics;
+import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


andyHa (approved):

src/main/java/sirius/db/mongo/Mango.java
```diff
@@ -160,7 +162,19 @@ private void writeField(MongoEntity entity, Updater updater, Property property)
      * @return <tt>true</tt> if the given value is a default value, <tt>false</tt> otherwise
      */
     private boolean isDefaultValue(Object valueForDatasource) {
-        return valueForDatasource == null || Boolean.FALSE.equals(valueForDatasource);
+        if (valueForDatasource == null) {
+            return true;
+        }
+
+        if (valueForDatasource instanceof List && ((List<?>) valueForDatasource).isEmpty()) {
```
qw3ry, line 169:
```suggestion
        if (valueForDatasource instanceof Collection && ((Collection<?>) valueForDatasource).isEmpty()) {
```
... that is, if Mongo supports collections other than lists. If this is not the case, it does not matter, obviously

qw3ry (approved):

src/main/java/sirius/db/mongo/Mango.java
idlira: Actually is not about mongo, but the object underlying our properties.

fhaScireum (approved):

cschierle (approved):

sabieber (approved):

PR #420, opened 2021-05-05
Permits to span an ELasticQuery across multiple indices.

andyHa:


sabieber (approved):

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -274,6 +283,26 @@ public ElasticQuery(EntityDescriptor descriptor, LowLevelClient client) {
         return list;
     }
 
+    /**
+     * Spans the query over the given additional indices.
+     * <p>
+     * Note that the given indices are added to the main index / descriptor which is arleady present. Also, not
```
mkeckmkeck, line 289:
```suggestion
     * Note that the given indices are added to the main index / descriptor which is already present. Also, not
```

mkeckmkeck (approved):
typo

PR #423, opened 2021-05-19
SkipDefaultValue: Improve handling/support and edge cases

sabieber:
## Considers nullability before actually checking if property changed

This is necessary as the annotation `SkipDefaultValue` allows to omit storing the default values of properties in the database (which means in MongoDB the field is actually completely missing from the stored object). When a field has any other default value than null isChanged would always return true before, as the equality check was comparing "null" (as the property is not in the persistedData map) against the default value of the entity (for example Amount.NOTHING).

We now utilize the `isConsideredNull` method of the property which effectively normalizes null and Amount.NOTHING to the same thing and compare them before running the equalsFunction.

## Considers implicit default value of primitive properties when checking if property changed

This is necessary as previously a null value from the datasource was compared against the initial value of the primitive field, which would not match up.

**Note:** This could be a behaviour change even when not using the `SkipDefaultValue`, but unlikely as cases where a primitive type is used and null is written to the DB should not occur.

Example:
Field: boolean property
Value from Datasource: null
Value in Entity: false

## Considers explicit default value via annotation when checking if property changed

This is changed to broaden the use cases where `SkipDefaultValue` can be used and should have absolutely no impact when not using the annotation.
Best to check the test case and test entity to see the intended use case in action.

Fixes: SIRI-557


idlira (approved):

cschierle (approved):

fhaScireum (approved):

mkeckmkeck (approved):

meg1502 (approved):

src/main/java/sirius/db/mixing/EntityDescriptor.java
```diff
@@ -441,7 +442,17 @@ public boolean isChanged(BaseEntity<?> entity, Property property) {
...
+        Object persistedValue = entity.persistedData.get(property);
+        Object newValue = property.getValue(entity);
+        if (property.isConsideredNull(persistedValue) && property.isConsideredNull(newValue)) {
+            return false;
+        }
+        if (property.isAnnotationPresent(SkipDefaultValue.class)
+            && property.isConsideredNull(persistedValue)
+            && Objects.equals(property.defaultValue, newValue)) {
+            return false;
+        }
```
andyHa, line 454:
```suggestion
        // The following is used to properly handle missing values in the database.
        // If, i.e. a boolean property has a "SkipDefaultValue" with the default being "false",
        // the persisted value will be "null" (as it wasn't stored at all) but the newValue will
        // be false (the default). In this case, this still doesn't count as a changed value, as
        // this way solely a storage optimization.
        if (property.isAnnotationPresent(SkipDefaultValue.class)
            && property.isConsideredNull(persistedValue)
            && Objects.equals(property.defaultValue, newValue)) {
            return false;
        }
```

PR #424, opened 2021-05-27
fix skipping entities for iterateBlockwiseByPaging by limit with skip

ymo-sci:
if there is no limit but we want to skip items, we have to set a limit for work with correct SQL Syntax like `SELECT ... ORDER BY... LIMIT 5,10`

### DIFF
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -822,11 +822,13 @@ public class SmartQuery<E extends SQLEntity> extends Query<SmartQuery<E>, E, SQL
     }
 
     private void limit(Compiler compiler) {
-        if (limit > 0 && db.hasCapability(Capability.LIMIT)) {
+        //if a skip value is set but no limit, we have to set a limit > 0 to make limit working.
+        int effectiveLimit = limit == 0 && skip > 0 ? Integer.MAX_VALUE : limit;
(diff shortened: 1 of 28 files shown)


src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -822,11 +822,13 @@ private void orderBy(Compiler compiler) {
     }
 
     private void limit(Compiler compiler) {
-        if (limit > 0 && db.hasCapability(Capability.LIMIT)) {
+        //if a skip value is set but no limit, we have to set a limit > 0 to make limit working.
+        int effectiveLimit = limit == 0 && skip > 0 ? Integer.MAX_VALUE : limit;
```
mkeckmkeck, line 826: This would always set a limit. 
Refactor it a bit more verbose. Maybe to check first for capability, then for skip (bring in effective limit only there) and then for limit?

mkeckmkeck (changes requested):

src/main/java/sirius/db/jdbc/SmartQuery.java
mkeckmkeck: Code should work. But the ternary with the and condition is maybe hard to read.

mkeckmkeck (approved):

qw3ry (approved):

andyHa (approved):

idlira (approved):

PR #426, opened 2021-06-02
Allow row values in OMA-SQL Queries and improve the iterate blockwise with them

qw3ry:
SIRI-357

# Improve the iterate blockwise using row values

The skip/limit construct is implemented rather slow in the database.
Because we know the value sorted right above the one we want to continue with, we can instead use this value, in conjunction with a total ordering to remove the entities already read from the result set.
This can result in a significant speedup, especially if you have the right indices, and even more so when there are expensive joins involved.

To implement this without too much hassle, we need the ability to use row values:

# Allow the usage of row values in sql constraints

A row value is what you'd call a tuple, and you can use it in SQL constraints like so:
```
... WHERE (column1, column2) > (5, "foo") ...
```
which is equivalent to
```
... WHERE column1 > 5 OR (column1 = 5 AND column2 > "foo") ...
```

A RowValue entity can contain both values and mappings, and you can also mix them.
As a drive-by, you can now also use them to compare columns with each other:
```
oma.select(Foo.class).where(OMA.FILTERS.eq(new RowValue(Foo.COLUMN1), new RowValue(Foo.COLUMN2)))
```
compiling to
```
... WHERE column1 = column2 ...
```
which might help with some more involved queries

### DIFF
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -305,22 +305,22 @@ public class SmartQuery<E extends SQLEntity> extends Query<SmartQuery<E>, E, SQL
         }
 
         // create and install the filter
-        Object[] lhs = new Object[query.orderBys.size()];
-        Object[] rhs = new Object[query.orderBys.size()];
+        Object[] leftHandSide = new Object[query.orderBys.size()];
+        Object[] rightHandSide = new Object[query.orderBys.size()];
         for (int i = 0; i < query.orderBys.size(); i++) {
             Mapping mapping = query.orderBys.get(i).getFirst();
             Object value = lastValue.getDescriptor().getProperty(mapping).getValue(lastValue);
             if (query.orderBys.get(i).getSecond().booleanValue()) {
                 // the order by is ascending -> COLUMN > lastvalue.column
-                lhs[i] = mapping;
-                rhs[i] = value;
+                leftHandSide[i] = mapping;
+                rightHandSide[i] = value;
             } else {
                 // lastvalue.column > COLUMN
-                rhs[i] = mapping;
-                lhs[i] = value;
+                rightHandSide[i] = mapping;
+                leftHandSide[i] = value;
             }
         }
-        return query.where(OMA.FILTERS.gt(new RowValue(lhs), new RowValue(rhs)));
+        return query.where(OMA.FILTERS.gt(new RowValue(leftHandSide), new RowValue(rightHandSide)));
     }
 
     @Override
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -8,7 +8,9 @@
 
 package sirius.db.jdbc;
 
+import sirius.db.jdbc.constraints.RowValue;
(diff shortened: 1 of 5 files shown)


andyHa (approved):

src/main/java/sirius/db/jdbc/constraints/RowValue.java
```diff
@@ -0,0 +1,80 @@
...
+ * An n-tuple of normal values that can be used in the OMA {@link sirius.db.jdbc.OMA#FILTERS filters}.
+ * <p>
+ * Use {@link Mapping} objects to indicate a column; every other object will be treated as an actual value.
+ * Usage example:<br>
+ * {@code
+ * oma.select(Entity.class).where(OMA.FILTERS.gte(new RowValue(Entity.COLUMN1, Entity.COLUMN2), new RowValue(4, "Test")))
+ * }
+ * <br>will compile to<br>
+ * {@code
+ * SELECT * FROM Entity e1 WHERE (e1.column1, e2.column2) >= (4, "Test")
```
qw3ry, line 33: Gibts was vergleichbares für Mongo/Elastic? Weil dann könnte man das ja gleich für alle machen.

mkeckmkeck (approved):

src/main/java/sirius/db/jdbc/constraints/FieldOperator.java
```diff
@@ -9,32 +9,37 @@
...
 /**
  * Represents a simple field operator as constraint.
  */
 class FieldOperator extends SQLConstraint {
 
-    private Mapping field;
-    private Object value;
-    private String op;
+    private final RowValue lhs;
+    private final RowValue rhs;
```
sabieber, line 23: Optional: Können wir da evtl deskriptivere Feldnamen etc wählen. Also ich vermute mal das steht für left und right hand side oder? Aber so richtig sicher ist man sich da dann nach ner Weile evtl net mehr wenn man da in nem halben Jahr was dran basteln muss.

src/main/java/sirius/db/jdbc/constraints/SQLFilterFactory.java
```diff
@@ -32,22 +32,96 @@ protected Object customTransform(Object value) {
...
+    }
+
+    /**
+     * Represents {@code lhs = rhs} as constraint
+     *
+     * @param lhs the first row value
+     * @param rhs the second row value
+     * @return the generated constraint
+     */
+    public SQLConstraint eq(RowValue lhs, RowValue rhs) {
```
sabieber, line 45: Optional: hier auch :)

src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -250,102 +256,71 @@ public void iterateBlockwise(Predicate<E> handler) {
...
+        // no last value => we query everything
+        if (lastValue == null) {
+            return query;
+        }
 
-                return true;
-            });
+        // create and install the filter
+        Object[] lhs = new Object[query.orderBys.size()];
+        Object[] rhs = new Object[query.orderBys.size()];
```
sabieber, line 309: Optional: hier auch

sabieber (approved):

ymo-sci (approved):

PR #434, opened 2021-06-30
Support `stream()` to execute queries, and fix SmartQuery's `iterateBlockwise[All]` to respect `limit` and `skip`

qw3ry:


src/main/java/sirius/db/mixing/query/BaseQuery.java
```diff
@@ -139,6 +140,20 @@ public Q fail() {
      */
     public abstract void iterate(Predicate<E> handler);
 
+    /**
+     * Executes the query and returnes the matched entities using the stream interface.
```
idlira, line 144:
```suggestion
     * Executes the query and returns the matched entities using the stream interface.
```

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -463,6 +465,17 @@ public ElasticQuery(EntityDescriptor descriptor, LowLevelClient client) {
         return this;
     }
 
+    @Override
+    public Stream<E> stream() {
+        // just in case no absolute ordering is present
+        orderAsc(Mapping.named(KEY_DOC_ID));
```
andyHa, line 471: check if orderings are present? this might kill an index if it is appended to an existing ordering (?)

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -1382,4 +1299,75 @@ public void truncate() {
     public String toString() {
         return descriptor.getType() + ": " + buildPayload();
     }
+
+    private class ElasticScrollingSpliterator extends PullBasedSpliterator<E> {
+        long lastScroll = 0;
```
andyHa, line 1304: private long ...

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -1382,4 +1299,75 @@ public void truncate() {
...
+        String scrollId = null;
```
andyHa, line 1305: private ...

src/main/java/sirius/db/util/AutoClosingStream.java
```diff
@@ -0,0 +1,289 @@
...
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+/**
+ * A stream wrapper calling {@link #close()} after any terminal operation.
+ *
+ * @param <T> the type of the stream elements
+ */
+public record AutoClosingStream<T>(Stream<T> delegate) implements Stream<T> {
```
andyHa, line 40: whats the benefit of making this a record?

qw3ry: There is none, its just recommended by IntelliJ

There is one tiny detail you cannot achieve without a record, which is that the `delegate` is truely constant this way. If you "just" declare it `final`, you can still modify it via Reflection. But I would not consider this a real "benefit".

src/main/java/sirius/db/es/ElasticQuery.java
qw3ry: Why would it kill an index? if you have `ORDER BY a, b, id`, and you have an index `a, b, c` this should be used by the database, as far as I know, and the "remaining" ordering by ID is done during evaluation.
However, if I leave the ID out, this might lead to inconsistencies if `ORDER BY a, b` is not a total ordering.
I do not know about the elastic implementation, though.

idlira (approved):

idlira (approved):
Since it touches an important part of the system including currently used methods, it's probably wise to merge it after the next ga release so products can give more testing time to it.

mkeckmkeck (approved):

PR #436, opened 2021-07-20
Prevents SQLSyntaxErrorException thrown as a result of `SmartQuery#count` and `SmartQuery#fields` combination

marvinwoop:
`COUNT` won't take multiple arguments unless it's a `COUNT(DISTINCT ...)` expression. Multiple fields in combination with `SmartQuery#count` will now result in `COUNT(*)` to prevent a SQLSyntaxErrorException.

### DIFF
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -808,11 +808,18 @@ public class SmartQuery<E extends SQLEntity> extends Query<SmartQuery<E>, E, SQL
         Compiler c = new Compiler(descriptor);
         c.getSELECTBuilder().append("SELECT COUNT(");
 
-        if (!fields.isEmpty() && distinct) {
-            c.getSELECTBuilder().append("DISTINCT ");
-            appendFieldList(c, false);
-        } else if (fields.stream().count() == 1) {
-            appendFieldList(c, false);
+        if (fields.size() > 1) {
+            if (distinct) {
+                c.getSELECTBuilder().append("DISTINCT ");
+                appendFieldList(c, false);
+            } else {
+                throw Exceptions.createHandled()
+                                .to(OMA.LOG)
+                                .withSystemErrorMessage(
+                                        "Only use multiple arguments in 'fields' and 'count' in "
+                                        + "combination with the 'distinct' statement")
+                                .handle();
+            }
         } else {
             c.getSELECTBuilder().append("*");
         }
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -806,16 +806,18 @@ public class SmartQuery<E extends SQLEntity> extends Query<SmartQuery<E>, E, SQL
 
     private Compiler selectCount() {
         Compiler c = new Compiler(descriptor);
-        if (!fields.isEmpty()) {
-            c.getSELECTBuilder().append("SELECT COUNT(");
-            if (distinct) {
-                c.getSELECTBuilder().append("DISTINCT");
-            }
+        c.getSELECTBuilder().append("SELECT COUNT(");
+
+        if (!fields.isEmpty() && distinct) {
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -806,16 +806,18 @@ private void appendToSELECT(Compiler c,
...
-            if (distinct) {
-                c.getSELECTBuilder().append("DISTINCT");
-            }
+        c.getSELECTBuilder().append("SELECT COUNT(");
+
+        if (!fields.isEmpty() && distinct) {
+            c.getSELECTBuilder().append("DISTINCT ");
+            appendFieldList(c, false);
+        } else if (fields.stream().count() == 1) {
             appendFieldList(c, false);
```
qw3ry, line 815: `SELECT count(id)` should be equivalent to `SELECT count(*)`, right? Then I'd prefer to just use `count(*)` in all cases

src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -806,16 +806,18 @@ private void appendToSELECT(Compiler c,
...
-            c.getSELECTBuilder().append("SELECT COUNT(");
-            if (distinct) {
-                c.getSELECTBuilder().append("DISTINCT");
-            }
+        c.getSELECTBuilder().append("SELECT COUNT(");
+
+        if (!fields.isEmpty() && distinct) {
+            c.getSELECTBuilder().append("DISTINCT ");
+            appendFieldList(c, false);
+        } else if (fields.stream().count() == 1) {
```
andyHa, line 814: Maybe throw an exception for multiple fields without distinct?

andyHa (approved):

src/main/java/sirius/db/jdbc/SmartQuery.java
qw3ry: https://www.w3resource.com/sql/aggregate-functions/count-function.php nope, small difference

andyHa: Or throw, as the programmer probably expected another outcome when specifying a field? like forgot to add "distinct"

qw3ry (approved):

src/main/java/sirius/db/jdbc/SmartQuery.java
marvinwoop: `SELECT COUNT(id)` doesn't take `null` values into account while `SELECT COUNT(*)` considers all fields of the row in it's calculation. It's possible that a count on id returns a different (lower) number of rows than a count on *. I'd therefore recommend to distinguish between those two statements.

ymo-sci (approved):

src/main/java/sirius/db/jdbc/SmartQuery.java
qw3ry:
```suggestion
        } else if (fields.size() == 1) {
```

marvinwoop: As discussed internally, we now treat a single field like `COUNT(*)` and multiple fields in combination with count and without distinct will throw an exception because we don't want the behaviour to appear that "magical". Fields in parent classes can be removed by calling something like `getQuery().fields().count()`.

mkeckmkeck (approved):

src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -806,16 +806,25 @@ private void appendToSELECT(Compiler c,
 
     private Compiler selectCount() {
         Compiler c = new Compiler(descriptor);
-        if (!fields.isEmpty()) {
-            c.getSELECTBuilder().append("SELECT COUNT(");
+        c.getSELECTBuilder().append("SELECT COUNT(");
```
sabieber, line 809: Optional: Rename c to compiler

sabieber: Optional: Extract getSELECTBUILDER as variable

sabieber (approved):

PR #439, opened 2021-08-19
BREAKING: Gives all properties a default value determined by their initial value

meg1502:
### BREAKING: Gives all properties a default value determined by their initial value

The biggest change from this is that all primitive number fields have a default value of '0' if not initialized with another number.
Previously they produced an IllegalArgumentException when they had no default value.
This means, for non-nullable fields that should not have a default value but should produce an error when left empty, a non-primitve datatype should be used. Then they will be properly transformed from import/webcontext to null and produce an error on save.

Please note that for primitive booleans this was already implemented, but if you used the default value directly, the actual default value has changed from '0' to 'false'.

### BREAKING:Changes Property#defaultValue type from String to Value

Instead of a string that doubles as a value used for the column default value, the defaultValue will now hold an Object that represents the property wrapped in a Value, or Value.EMPTY if no default Value is present. JDBC Column default value can be retrieved via a new method.

To make it more universal, DefaultValue Annotations are now transformed like machine input (Property#transformValueFromImport), so existing DefaultValue annotations might need to be adjusted, especially times,dates and booleans.

### DIFF
--- a/src/main/java/sirius/db/mixing/Property.java
+++ b/src/main/java/sirius/db/mixing/Property.java
@@ -112,9 +112,11 @@ public abstract class Property extends Composable {
     protected Field field;
 
     /**
-     * Contains a string representation of the default value for the column
+     * Contains the default value of this property
+     * <p>
+     * For the string representation that can be used in JDBC statements see {@link #getColumnDefaultValue()}
      */
-    protected String defaultValue;
+    protected Value defaultValue = Value.EMPTY;
 
     /**
      * Contains the length of this property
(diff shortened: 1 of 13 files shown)


src/main/java/sirius/db/mixing/Property.java
```diff
@@ -162,6 +162,11 @@ protected void determineDefaultValue() {
         DefaultValue dv = field.getAnnotation(DefaultValue.class);
         if (dv != null) {
             this.defaultValue = dv.value();
+        } else {
+            Object initalValue = getValue(getDescriptor().getReferenceInstance());
```
idlira, line 168: typo: initialValue

src/test/java/sirius/db/mixing/properties/DefaultValuesSpec.groovy
```diff
@@ -0,0 +1,126 @@
...
+    @Part
+    private static Mixing mixing
+
+    def "the default values are properly initialized"() {
+        expect:
+        mixing.getDescriptor(MongoDefaultValuesEntity.class).findProperty(propertyName).getDefaultValue() ==
+                excpectedDefault
+
+        where:
+        propertyName            | excpectedDefault
```
idlira, line 30: typo: expected

idlira (changes requested):

idlira (approved):

qw3ry (approved):
- There seems to be quite a lot to migrate. A short "Migration Guide" would be helpful
- Would `@DefaultValue(Value.EMPTY) int x` also work to get a non-nullable field without default value?

meg1502:
- I can write a short guide, it isn't that much as the Default Value Annotation isn't that much used, the main thing will be ensuring all non-nullable fields that should report an error when saving instead of applying a default are non-primitive
- Then we would be back to square 1, the main point of this change is that primitve fields are no longer used for this case, as primitive fields will always produce an illegalArgumentException instead of the "can't be null" error, as the field need to be able to hold the null value between transforming from user input/import and the saving of the entity where the nullability check happens.

andyHa (approved):

andyHa:
Are any @DefaultValue annotations present anyway? It would be much safer and more consistent to simply initialize the
Java field anyway(?)

PR #443, opened 2021-08-25
Fix the default value of amount properties

qw3ry:


src/main/java/sirius/db/mixing/properties/AmountProperty.java
```diff
@@ -111,14 +114,21 @@ protected Object transformValueFromImport(Value value) {
...
     protected Object transformToJDBC(Object object) {
-        return object == null || ((Amount) object).isEmpty() ? null : ((Amount) object).getAmount();
+        Amount amount = (Amount) object;
+        if (amount == null || amount.isEmpty()) {
+            return null;
+        }
+        NumberFormat format = getAnnotation(Numeric.class).map(numeric -> {
+            return new NumberFormat(numeric.scale(), RoundingMode.HALF_UP, NLS.getMachineFormatSymbols(), false, null);
+        }).orElse(NumberFormat.MACHINE_THREE_DECIMAL_PLACES);
+        return amount.toString(format);
```
meg1502, line 124: Are you sure we want to do this here, as this also influences the transformation of actual values, which were fine before, i assume ?

Maybe a better way would be to overwrite getColumnDefaulValue in AmountProperty

andyHa, line 124: why would we choose a string over a BigDecimal? many JDBC drivers (esp. IBM i5 expect a BigDecimal here).

Also, rounding within a coercion method sounds like a risky idea

qw3ry: In general, this shouldn't be a problem - the values cannot be stored with higher precision in the DB. But as there might be some edge cases, I agree.

meg1502 (approved):

PR #448, opened 2021-09-06
Builds an 'Entity cannot be saved' message

idlira:
`EntityDescriptior.createCannotSaveMessage(MyEntity.class, message)`

Fixes: OX-7244


sabieber (approved):

src/main/java/sirius/db/mixing/BaseEntity.java
```diff
@@ -66,12 +66,43 @@
...
+     * This message is mostly used as a consequence describing why an error happened.
+     * For instance, when an entity's required field is not populated, a standard message like this will be logged:
+     * {@code Field 'MyField' is required.}
+     * As to would be rather nice to tell the end user that an entity couldn't be saved due to the error.
+     * {@code MyEntity cannot be saved. Field 'MyField' is required.}
+     *
+     * @param clazz the entity class to extract its label
+     * @return the formatted message
+     */
+    public static String createCannotSaveMessage(@Nonnull Class<?> clazz) {
```
meg1502, line 87: Do we really need the static implementation?

meg1502 (approved):

src/main/java/sirius/db/mixing/BaseEntity.java
```diff
@@ -33,7 +33,7 @@
  * result in the equivalent properties required to store the fields declared there. Still inheritance might be
  * useful and is fully supported for both, entities and composites.
  * <p>
- * What is not supported, is merging distinct subclasses into one table or other weired inheritance methods. Therefore
+ * What is not supported, is merging distinct subclasses into one table or other weired inheritance methods. Therefore,
```
cschierle, line 36:
```suggestion
 * What is not supported, is merging distinct subclasses into one table or other weird inheritance methods. Therefore,
```

src/main/java/sirius/db/mixing/BaseEntity.java
idlira: We do, because we must build this message for logging even before an entity even comes to exist. For example in context of ImportHandlers.

src/main/java/sirius/db/mixing/BaseEntity.java
```diff
@@ -66,12 +66,43 @@
...
         return mixing.getDescriptor(getClass());
     }
 
+    /**
+     * Builds a standard message that the given entity cannot be saved.
+     * <p>
+     * This message is mostly used as a consequence describing why an error happened.
+     * For instance, when an entity's required field is not populated, a standard message like this will be logged:
+     * {@code Field 'MyField' is required.}
+     * As to would be rather nice to tell the end user that an entity couldn't be saved due to the error.
```
cschierle, line 81: I don't understand really understand this explanation :sweat_smile:  Maybe try to make it a bit clearer...

src/main/java/sirius/db/mixing/BaseEntity.java
```diff
@@ -66,12 +66,43 @@
...
+        return NLS.fmtr("BaseEntity.cannotSaveEntity").set("entity", mixing.getDescriptor(clazz).getLabel()).format();
+    }
+
+    /**
+     * Builds a standard message that the entity cannot be saved.
+     * <p>
+     * This message is mostly used as a consequence describing why an error happened.
+     * For instance, when an entity's required field is not populated, a standard message like this will be logged:
+     * {@code Field 'MyField' is required.}
+     * As to would be rather nice to tell the end user that an entity couldn't be saved due to the error.
```
cschierle, line 97: s.a.

src/main/java/sirius/db/mixing/BaseEntity.java
andyHa: move to descriptor?

src/main/resources/db_de.properties
```diff
@@ -1,3 +1,4 @@
+BaseEntity.cannotSaveEntity = ${entity} wurde nicht gespeichert.
```
andyHa, line 1:
```suggestion
BaseEntity.cannotSaveEntity = ${entity} kann nicht gespeichert werden:
```

src/main/java/sirius/db/mixing/BaseEntity.java
idlira: Where, there is an example! Not sure how better I could describe it 🤔

src/main/resources/db_en.properties
```diff
@@ -1,3 +1,4 @@
+BaseEntity.cannotSaveEntity = ${entity} cannot be saved.
```
andyHa, line 1:
```suggestion
BaseEntity.cannotSaveEntity = ${entity} cannot be saved:
```

src/main/resources/db_de.properties
idlira: Actually, do we want to add punctuation at all? Would limit it to always be placed at the beginning.

src/main/java/sirius/db/mixing/BaseEntity.java
cschierle: I mean which of the 2 variants is now actually logged?
It just says that "it would be nice to tell the entity", but do we always do that now, or are both variants still possible?

sabieber (approved):

src/main/java/sirius/db/mixing/EntityDescriptor.java
```diff
@@ -215,6 +219,39 @@ protected EntityDescriptor(Class<?> type) {
...
+     * This message is mostly used as a consequence describing why an error happened.
+     * For instance, when an entity's required field is not populated, a standard message like this will be logged:
+     * {@code Field 'MyField' is required.}
+     * As to would be rather nice to tell the end user that an entity couldn't be saved due to the error.
+     * {@code MyEntity cannot be saved: Field 'MyField' is required.}
+     *
+     * @param clazz the entity class to extract its label
+     * @return the formatted message
+     */
+    public static String createCannotSaveMessage(@Nonnull Class<?> clazz) {
```
andyHa, line 234: do we need this? this is just a shortcut for mixing.getDescriptor(clazz).createCannotSaveMessage()

idlira: I think is useful boilerplate if you have an entity at hand. Otherwise you must (at least) initialize a part...

src/main/java/sirius/db/mixing/EntityDescriptor.java
```diff
@@ -215,6 +219,39 @@ protected EntityDescriptor(Class<?> type) {
...
+    }
+
+    /**
+     * Builds a standard message that the given entity cannot be saved.
+     * <p>
+     * This message is mostly used as a consequence describing why an error happened.
+     * For instance, when an entity's required field is not populated, a standard message like this will be logged:
+     * {@code Field 'MyField' is required.}
+     * As to would be rather nice to tell the end user that an entity couldn't be saved due to the error.
+     * {@code MyEntity cannot be saved: Field 'MyField' is required.}
```
andyHa, line 247:
```suggestion
     * This message is mostly used as a consequence for describing why an error happened. Therefore, this will 
     * create message like <tt>Enttiy cannot be saved: Field-Error</tt> instead of logging an abitrary looking field 
     * error alone.
```

andyHa (approved):

mkeckmkeck (approved):

PR #454, opened 2021-09-14
Drops milliseconds when persisting LocalDateTime in MongoDB

idlira:
When storing `LocalDateTime` in MongoDB, it only stores up to 1 ms as precision, but in recent Java we have nanoseconds.
Therefore, fields of LocalDateTimeProperty were being identified as changed (especially TraceData.changedAt), when 2 updates were issued within 1 ms but nothing changed, resulting in the `Tried to update the changed entity ... but actually nothing was changed in the database!`
By using seconds we also keep it on-pair with how LocalDateTimeProperty are handled in OMA and ES.

We also adjust the isChanged method to only compare dates up to 1 second when comparing dates.

Fixes: OX-7281

### DIFF
--- a/src/main/java/sirius/db/mixing/properties/LocalDateTimeProperty.java
+++ b/src/main/java/sirius/db/mixing/properties/LocalDateTimeProperty.java
@@ -151,7 +151,7 @@ public class LocalDateTimeProperty extends Property implements ESPropertyInfo, S
      * Overrides the default behavior, as the initial value of a temporal property is not suited for a default.
      * <p>
      * The initial value will commonly be a temporal value and thus not a constant.
-     * Therefore we ignore the initial value here, and only check for a {@link DefaultValue} annotation on the field.
+     * Therefore, we ignore the initial value here, and only check for a {@link DefaultValue} annotation on the field.
      */
     @Override
     protected void determineDefaultValue() {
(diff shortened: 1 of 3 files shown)


sabieber (approved):

mkeckmkeck (approved):

andyHa (approved):

src/main/java/sirius/db/mixing/properties/LocalDateTimeProperty.java
```diff
@@ -124,11 +125,10 @@ protected Object transformToElastic(Object object) {
 
     @Override
     protected Object transformToMongo(Object object) {
-        if (!(object instanceof LocalDateTime)) {
-            return null;
+        if (object instanceof LocalDateTime localDateTime) {
+            return QueryBuilder.FILTERS.transform(localDateTime == null ? null : localDateTime.withNano(0));
```
andyHa, line 129: shouldn't we move this into the transform method ->
MongoFilterFactory.customTransform(Object value)

Becaus we shoud apply "withNano(0)" on all values to be compared against a stored values as otherwise the comparison might fail...

andyHa (changes requested):

sabieber (approved):

cschierle (approved):

andyHa (approved):

PR #455, opened 2021-09-21
Enables to remove an entity from a cache by only specifying its id.

fhaScireum:
This is useful in cases where two entities are related but changing one does not directly affect a persisted field of the other, but rather a transient field. Instead of modifying the transient field, we may want to load a clean instance of the other entity if it is needed again.

- Fixes: OX-7421

### DIFF
--- a/src/main/java/sirius/db/util/BaseEntityCache.java
+++ b/src/main/java/sirius/db/util/BaseEntityCache.java
@@ -16,6 +16,7 @@ import sirius.kernel.commons.Strings;
 import sirius.kernel.health.Exceptions;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.util.Optional;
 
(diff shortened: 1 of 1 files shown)


sabieber (approved):

src/main/java/sirius/db/util/BaseEntityCache.java
```diff
@@ -126,9 +127,20 @@ public E fetchRequired(@Nonnull BaseEntityRef<I, E> entityRef) {
...
             entityByIdCache.remove(entity.getIdAsString());
         }
     }
+
+    /**
+     * Purges the entity specified by its id from the cache.
+     *
+     * @param entityId the id of the entity to purge
+     */
+    public void remove(@Nullable String entityId) {
```
andyHa, line 141:
```suggestion
    public void removeById(@Nullable String entityId) {
```

idlira (approved):

mkeckmkeck (approved):

meg1502 (approved):

jmuscireum (approved):

PR #460, opened 2021-10-04
Kills collation handling for MongoDB.

andyHa:
Collations are very tricky to use properly and ruin almost any ad-hoc
query as indices need to be aware of them. Also, we ran into strange
bugs / shortcomings with index misses for prefix/$regex queries.

We therefore resort to a normalized sort column:
Strings.reduceCharacters(...).toLowerCase() for sorting and ignore
collations for everything else.

Fixes: SIRI-440

Note that all indices need to be re-created after this has been merged!

### DIFF
--- a/src/main/java/sirius/db/es/constraints/BoolQueryBuilder.java
+++ b/src/main/java/sirius/db/es/constraints/BoolQueryBuilder.java
@@ -14,6 +14,7 @@ import sirius.db.es.ElasticQuery;
 import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Strings;
 
+import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Predicate;
@@ -171,8 +172,9 @@ public class BoolQueryBuilder {
      *
      * @return the query as constraint
      */
-    @SuppressWarnings({"squid:MethodCyclomaticComplexity","java:S1168"})
-    @Explain("Splitting this method would most probably increase the complexity. Also, we return null to indicate that no filter has been generated.")
+    @SuppressWarnings("squid:MethodCyclomaticComplexity")
+    @Explain("Splitting this method would most probably increase the complexity.")
+    @Nullable
     public JSONObject build() {
         int filters = filter == null ? 0 : filter.size();
         int musts = must == null ? 0 : must.size();
--- a/src/main/java/sirius/db/es/constraints/BoolQueryBuilder.java
+++ b/src/main/java/sirius/db/es/constraints/BoolQueryBuilder.java
@@ -171,8 +171,8 @@ public class BoolQueryBuilder {
      *
      * @return the query as constraint
      */
-    @SuppressWarnings("squid:MethodCyclomaticComplexity")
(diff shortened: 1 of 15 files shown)


src/main/java/sirius/db/mixing/properties/MultiPointLocationProperty.java
```diff
@@ -101,18 +100,16 @@ public Object transformFromDatasource(Class<? extends BaseMapper<?, ?, ?>> mappe
...
-
         if (!object.isNull()) {
             Object coordinates = ((Document) object.get()).get("coordinates");
             if (coordinates instanceof List<?>) {
-                ((List<List<Double>>) coordinates).stream()
-                                                  .map(entry -> Tuple.create(entry.get(0), entry.get(1)))
-                                                  .collect(Lambdas.into(locations));
+                return ((List<List<Double>>) coordinates).stream()
+                                                         .map(entry -> Tuple.create(entry.get(0), entry.get(1)))
+                                                         .collect(Collectors.toList());
```
sabieber, line 108: This means the returned list can not be altered afterwards, right? Is that okay in this case?

sabieber (changes requested):

src/main/java/sirius/db/mixing/properties/MultiPointLocationProperty.java
andyHa: stream.toList() is immutable / stream.collect(Collectors.toList()) is an array list

sabieber: Ahhhh, thanks +1 good to know

src/main/java/sirius/db/mongo/SortField.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class SortField extends Composite {
+
+    /**
+     * Contains the effective field to sort by.
+     */
+    public static final Mapping SORT_FIELD = Mapping.named("sortField");
+    @NullAllowed
+    @SuppressWarnings("java:S1700")
+    @Explain("We actually to want the same name here.")
```
idlira, line 37:
```suggestion
    @Explain("We actually do want the same name here.")
```

src/main/java/sirius/db/es/constraints/BoolQueryBuilder.java
```diff
@@ -171,8 +171,8 @@ public BoolQueryBuilder named(String name) {
      *
      * @return the query as constraint
      */
-    @SuppressWarnings("squid:MethodCyclomaticComplexity")
-    @Explain("Splitting this method would most probably increase the complexity")
+    @SuppressWarnings({"squid:MethodCyclomaticComplexity","java:S1168"})
+    @Explain("Splitting this method would most probably increase the complexity. Also, we return null to indicate that no filter has been generated.")
```
idlira, line 175: add the @Nullable annotation since we return null?

idlira (changes requested):

idlira (approved):

sabieber (approved):

jmuscireum (approved):

mkeckmkeck (approved):

PR #462, opened 2021-10-06
Makes the  SortField normalization method public

idlira:
Fixes: OX-7542


src/main/java/sirius/db/mongo/SortField.java
```diff
@@ -50,6 +50,16 @@ public SortField(MongoEntity owner) {
         this.owner = owner;
     }
 
+    /**
+     * Normalizes the given text using {@link Strings#reduceCharacters(String)}, lower-casing it after.
+     *
+     * @param text the text to normalize
+     * @return the normalized text
```
cschierle, line 57: maybe also clarify this in the return JavaDoc
```suggestion
     * @return the normalized text in lower-case
```

qw3ry (approved):

meg1502 (approved):

PR #465, opened 2021-10-08
Mongo $size

jakobvogel:


src/main/java/sirius/db/mongo/constraints/MongoFilterFactory.java
```diff
@@ -278,4 +278,18 @@ public MongoConstraint text(String value) {
...
+     * Builds a constraint regarding the {@code $size} of a (list) mapping.
+     * <p>
+     * See <a href="https://docs.mongodb.com/manual/reference/operator/query/size/">the manual</a> for usage information
+     * and restrictions concerning the operator.
+     *
+     * @param mapping the mapping to filter
+     * @param value the target size to compare against
+     * @return the generated constraint
+     */
+    public MongoConstraint size(Mapping mapping, Integer value) {
```
andyHa, line 292: shouldn't we call this s.th. like "hasListSize" or the like?

And why is value nullable? what is the semantics of null here? Should be either an int or explained

andyHa (approved):

PR #466, opened 2021-10-12
Introduces onValidate in properties

idlira:
A Property now defines an onValidate method (empty by default), which can be overwritten if validation code can be written at property level

Fixes: SIRI-446

### DIFF
--- a/src/main/java/sirius/db/mixing/Property.java
+++ b/src/main/java/sirius/db/mixing/Property.java
@@ -293,7 +293,7 @@ public abstract class Property extends Composable {
     }
 
     /**
-     * Returns the full label or nme of the property which is shown in error messages etc.
+     * Returns the full label or name of the property which is shown in error messages etc.
      * <p>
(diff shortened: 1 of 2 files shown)


jmuscireum (approved):

src/main/java/sirius/db/mixing/Property.java
```diff
@@ -670,6 +671,17 @@ protected final void onBeforeSave(Object entity) {
...
+    /**
+     * Invoked during validation of an entity.
+     * <p>
+     * This method is intended to be overwritten with custom logic.
+     *
+     * @param entity             the entity to check
+     * @param validationConsumer the consumer collecting the validation messages
+     */
+    protected void onValidate(Object entity, Consumer<String> validationConsumer) {
+    }
```
sabieber, line 683: Doesnt SonarLint want a comment explaining why its empty here?

sabieber (approved):

PR #489, opened 2022-02-17
Query Tool Syntax

jakobvogel:
> Please review carefully, not just with a pinch but tons of salt! **If in doubt, reject this PR.** (Ideally, providing an alternative solution idea. 😇)

## TL;DR

Extends the query tool syntax to support queries such as `ABC.XYZ: 1234` not only for referenced properties, but also for maps.
**Read the remaining text as well!** 😛

## Motivation

MIO uses nested maps quite often, for instance for storing tags and references. Currently, such maps are not accessible via the query tool, and require direct database access for respective queries.

## Implementation

This PR changes two things:
1. Queries containing a _dot_ are now parsed as nested mapping if not processing an entity reference (rather than **null**).
2. _Elastic's_ implementation of `eqValue` has been extended to treat nested mappings with string value as map lookups. (🚨🚨🚨)

## Caveats

1. **OMA compatibility has not been checked!**
2. **Overwriting `eqValue` may have considerable side effects?**


sabieber (approved):

src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -373,13 +373,15 @@ private boolean continueToken(boolean inQuotes) {
...
-            return null;
+        // properly resolve reference properties
+        if (property instanceof BaseEntityRefProperty<?, ?, ?> baseEntityRefProperty) {
+            return resolveProperty(baseEntityRefProperty.getReferencedDescriptor(), splitPath.getSecond(), effectiveMapping);
         }
 
-        return resolveProperty(((BaseEntityRefProperty<?, ?, ?>) property).getReferencedDescriptor(),
-                               splitPath.getSecond(),
-                               effectiveMapping);
+        // reaching this point, we likely have steampunk database access with most complicated structures; just add the
```
andyHa, line 381: as far as I understand it, our most common use-case is indexing into a StringMap or into a NestedList? Couldn't we handle this case by simply returning Tuple(Mapping.named(splitPath.getSecond()), property);

And then overwrite the QueryCompiler.compileFieldEquals to also detect properties of this kind and then generate the proper constraint (as implemented above). This way, we woudln't have to change the FilterFactory at all, and we would be sort of sure, that we properly handle the query in a predictive way (or would throw an error otherwise)...

jakobvogel: I left the `join` as more elegant expression than `inner`; Mongo seems to handle it alright, and we have special processing for Elastic anyways, and the recursive structure occurs to match the DB structure more accurately. Correct me if I am wrong. 😅

sabieber (approved):

sabieber (approved):

src/main/java/sirius/db/es/constraints/ElasticQueryCompiler.java
```diff
@@ -49,4 +52,22 @@ protected ElasticConstraint compileSearchToken(Mapping field, QueryField.Mode mo
...
+        if (property instanceof BaseMapProperty) {
+            return Tuple.create(Mapping.named(nestedPath), property);
+        }
+
+        return null;
+    }
+
+    @Override
+    protected ElasticConstraint compileFieldEquals(Mapping field, Property property, FieldValue value) {
+        if (value.getValue() instanceof String stringValue && (property instanceof BaseMapProperty)) {
```
jakobvogel, line 67:
```suggestion
        if (value.getValue() instanceof String stringValue && property instanceof BaseMapProperty) {
```

Oder beides klammern?

src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -666,7 +678,14 @@ protected C compileNotEquals(Mapping field, Property property, FieldValue value)
...
      * @return a constraint matching the given parameters
      */
+    @SuppressWarnings("unchecked")
+    @Explain("We securely cast a FilterFactory<C> to special implementations where we are sure about the identity of C.")
     protected C compileFieldEquals(Mapping field, Property property, FieldValue value) {
+        if (factory instanceof ElasticFilterFactory elasticFilterFactory
+            && value.getValue() instanceof String stringValue
+            && (property instanceof StringMapProperty || property instanceof StringListMapProperty)) {
+            return (C) elasticFilterFactory.nestedMapContains(field.getParent(), field.getName(), stringValue);
+        }
```
jakobvogel, line 688: Brauchen wir nicht mehr?

PR #497, opened 2022-04-21
In ElasticSearch: Use new in field id in ElasticEntity 

ReadMails-ReallyFast:
For aggreations, sorting and scripting the usage of _id is depricated, a copy this field is needed to use for these tasks. see: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-id-field.html


fhaScireum (approved):

src/main/java/sirius/db/es/Elastic.java
```diff
@@ -327,10 +327,8 @@ protected void updateEntity(ElasticEntity entity, boolean force, EntityDescripto
     protected boolean toJSON(EntityDescriptor ed, ElasticEntity entity, JSONObject data) {
         boolean changed = false;
         for (Property p : ed.getProperties()) {
-            if (!ElasticEntity.ID.getName().equals(p.getName())) {
-                data.put(p.getPropertyName(), p.getValueForDatasource(Elastic.class, entity));
-                changed |= ed.isChanged(entity, p);
-            }
+            data.put(p.getPropertyName(), p.getValueForDatasource(Elastic.class, entity));
```
idlira, line 330: hm, this commit is from the other PR...

PR #500, opened 2022-05-03
Asserts modified fields

idlira:
Helper method used to track if a field changed (after insert), much likely in a BeforeSave handler.

Fixes: OX-8433

### DIFF
--- a/src/main/java/sirius/db/mixing/BaseEntity.java
+++ b/src/main/java/sirius/db/mixing/BaseEntity.java
@@ -187,7 +187,7 @@ public abstract class BaseEntity<I> extends Mixable implements Entity {
      *
      * @param field the field to check
      */
-    public void assertFieldChanged(Mapping field) {
+    public void assertFieldUnchanged(Mapping field) {
         if (!isNew() && isChanged(field)) {
             Property property = getDescriptor().getProperty(field);
             throw Exceptions.createHandled()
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/db/mixing/BaseEntity.java
```diff
@@ -180,6 +180,24 @@ private void throwEmptyFieldError(Mapping field, Property property) {
...
     }
 
+    /**
+     * Asserts that the given field remains unchanged.
+     * <p>
+     * This can be used for conditions where a field cannot be updated after creation.
+     *
+     * @param field the field to check
+     */
+    public void assertFieldChanged(Mapping field) {
```
andyHa, line 190:
```suggestion
    public void assertFieldUnchanged(Mapping field) {
```

sabieber, line 190: The method name does not describe the functionality, maybe something like `assertFieldUnchanged`?

sabieber: Maybe also `Mapping...` instead of `Mapping` to allow more use cases? But would need some small adaptions when generating the message

sabieber (changes requested):

sabieber (approved):

mkeckmkeck (approved):

src/main/java/sirius/db/mixing/BaseEntity.java
idlira: Dachte auch. Aber der `InvalidFieldException` erwartet genau ein Feld, sonst brauchen wir eine andere Exception.

sabieber: kk

cschierle (approved):

PR #506, opened 2022-06-05
Properly converts entity instances with an empty toString method.

andyHa:
This filter was only targeting empty IDs, therefore we switched the order
to always pass-through values of the expected entity type - even is their
toString representation is empty.


andyHa:
Better fix than:
* https://github.com/scireum/datahive/pull/610

mkeckmkeck (approved):

sabieber (approved):

idlira (approved):

jmuscireum (approved):

cschierle (approved):

PR #513, opened 2022-07-06
Feature/aha/dependency update

andyHa:


jakobvogel (approved):

pom.xml
```diff
@@ -3,6 +3,18 @@
...
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>17</source>
+                    <target>17</target>
```
sabieber: Not 18? Why is this even defined here? This should come from sirius parent

sabieber (changes requested):

pom.xml
```diff
@@ -83,7 +83,7 @@
         <dependency>
             <groupId>ru.yandex.clickhouse</groupId>
             <artifactId>clickhouse-jdbc</artifactId>
-            <version>0.2.6</version>
+            <version>0.3.2</version>
```
andyHa, line 86:
```suggestion
            <version>0.2.6</version>
```

pom.xml
andyHa: This was an accident all by itself and has been removed again....

idlira (approved):

PR #519, opened 2022-09-05
SIRI-626 stream blockwise bug

qw3ry:
Please look carefully to avoid follow-up bugs

### DIFF
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -337,6 +337,7 @@ public class SmartQuery<E extends SQLEntity> extends Query<SmartQuery<E>, E, SQL
                     missingOrderBys.forEach(adjusted::orderAsc);
                 }
             } else {
+                // we are not DISTINCT, so we can easily guarantee an absolute ordering using the ID
                 adjusted.orderAsc(BaseEntity.ID);
             }
 
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -41,6 +41,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -320,8 +318,44 @@ protected Iterator<E> pullNextBlock() {
...
+            }
+
+            // we need to guarantee an absolute ordering
+            if (adjusted.distinct) {
+                // we have distinct fields, so we can easily create an absolute ordering, if it's not already there
+                if (!new HashSet<>(Tuple.firsts(adjusted.orderBys)).containsAll(adjusted.fields)) {
+                    Set<Mapping> missingOrderBys = new HashSet<>(adjusted.fields);
+                    Tuple.firsts(orderBys).forEach(missingOrderBys::remove);
+                    missingOrderBys.forEach(adjusted::orderAsc);
+                }
```
andyHa, line 338: couldn't this be fixed in a single stream pipeline - should kinda be possible (?)

src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -320,8 +318,44 @@ protected Iterator<E> pullNextBlock() {
...
+
+            if (!adjusted.distinct && !fields.isEmpty()) {
+                // We SELECT a subset of the columns to optimize the network bandwidth.
+                // When pulling the next block, we need to continue exactly where we left of, so we need to SELECT
+                // at least all the fields from the ORDER BY clause.
+                Set<Mapping> allFields = new HashSet<>(adjusted.fields);
+                if (!allFields.containsAll(Tuple.firsts(adjusted.orderBys))) {
+                    allFields.addAll(Tuple.firsts(adjusted.orderBys));
+                }
+                adjusted.fields(allFields.toArray(Mapping[]::new));
```
andyHa, line 351: shoudln't this be a pre-allocated 0-sized array?

src/main/java/sirius/db/jdbc/SmartQuery.java
andyHa: is this ordering guaranteed to be absolute? shouldn't the id still be there just in case?

src/main/java/sirius/db/jdbc/SmartQuery.java
andyHa: don't we still need to add ID as ordering to guarantee stable sorting?

src/main/java/sirius/db/jdbc/SmartQuery.java
qw3ry: The query is `distinct`, so it should suffice to ORDER BY all the selected fields

src/main/java/sirius/db/jdbc/SmartQuery.java
qw3ry: The ID is added in line 340. Do you prefer this more explicitly?

qw3ry: Since JDK11, there is [`Collection#toArray(IntFunction<T[]>)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Collection.html#toArray(java.util.function.IntFunction)), which takes a reference to an array constructor instead. The only purpose of the generator function / 0-sized array / pre-sized array is, that the method can somehow construct an array of the correct type. Using the generator function seems to be the most ideomatic way to do this.

mkeckmkeck (approved):

andyHa (approved):

sabieber (approved):

idlira (approved):

PR #535, opened 2022-11-15
Improve the SmartQuery#streamBlockwise

qw3ry:
Previously, we required the querys ORDER BY clauses did not contain JOINed mappings. Now we can support these

### DIFF
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -392,7 +392,7 @@ public class SmartQuery<E extends SQLEntity> extends Query<SmartQuery<E>, E, SQL
             return parent.getDescriptor().getProperty(mapping.getName()).getValue(parent);
         }
 
-        private static BaseEntity<?> findParent(Mapping mapping, BaseEntity<?> entity) {
+        private BaseEntity<?> findParent(Mapping mapping, BaseEntity<?> entity) {
             if (mapping.getParent() != null) {
                 BaseEntity<?> parentEntity = findParent(mapping.getParent(), entity);
                 if (parentEntity.getDescriptor()
--- a/src/main/java/sirius/db/jdbc/SmartQuery.java
+++ b/src/main/java/sirius/db/jdbc/SmartQuery.java
@@ -14,9 +14,12 @@ import sirius.db.mixing.BaseEntity;
 import sirius.db.mixing.BaseMapper;
 import sirius.db.mixing.EntityDescriptor;
 import sirius.db.mixing.Mapping;
+import sirius.db.mixing.Property;
+import sirius.db.mixing.properties.BaseEntityRefProperty;
 import sirius.db.mixing.properties.SQLEntityRefProperty;
 import sirius.db.mixing.query.Query;
 import sirius.db.mixing.query.constraints.FilterFactory;
+import sirius.db.mixing.types.BaseEntityRef;
 import sirius.kernel.async.TaskContext;
 import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Limit;
@@ -370,7 +373,7 @@ public class SmartQuery<E extends SQLEntity> extends Query<SmartQuery<E>, E, SQL
             for (Tuple<Mapping, Boolean> sorting : effectiveQuery.orderBys) {
                 Mapping sortColumn = sorting.getFirst();
                 boolean sortAscending = sorting.getSecond().booleanValue();
-                Object value = lastValue.getDescriptor().getPropertyValue(sortColumn, lastValue, false);
+                Object value = getPropertyValue(lastValue.getDescriptor(), sortColumn, lastValue);
                 if (sortAscending) {
                     // the order by is ascending -> COLUMN > lastvalue.column
                     leftHandSide.addComponent(sortColumn);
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/db/mixing/EntityDescriptor.java
```diff
@@ -1045,6 +1047,35 @@ public Property getProperty(String property) {
...
+    public Object getPropertyValue(Mapping mapping, Object entity, boolean resolveJoins) {
+        if (mapping.getParent() != null) {
+            Object parent = getPropertyValue(mapping.getParent(), entity, resolveJoins);
+            return mixing.getDescriptor(parent.getClass())
+                         .getPropertyValue(Mapping.named(mapping.getName()), parent, resolveJoins);
+        }
+        Property property = getProperty(mapping.getName());
+        if (property instanceof BaseEntityRefProperty<?, ?, ?> ref) {
+            BaseEntityRef<?, ?> entityRef = ref.getEntityRef(entity);
+            return resolveJoins ? entityRef.fetchValue() : entityRef.getValueIfPresent().orElseThrow();
```
mkeckmkeck, line 1074: Throw exception with specific cause saying query not build up properly as joined mapping not resolved

qw3ry: It's not by design a query that was not properly joined - the method can be used from anywhere, not just `#streamBlockwise`. However, a sensible error message does indeed make sense here.

mkeckmkeck (approved):

src/main/java/sirius/db/mixing/properties/BaseEntityRefProperty.java
```diff
@@ -142,8 +142,16 @@ protected Object transformToDatasource(Class<? extends BaseMapper<?, ?, ?>> mapp
      */
     protected abstract Optional<E> find(Class<E> type, Value value);
 
+    /**
+     * Resolve the referenced EntityRef
```
jakobvogel, line 146: The referenced entity reference? Please clarify 😅

```suggestion
     * Resolves the referenced entity.
```

jakobvogel (approved):

src/main/java/sirius/db/mixing/EntityDescriptor.java
```diff
@@ -1045,6 +1047,40 @@ public Property getProperty(String property) {
...
+                return new IllegalArgumentException(Strings.apply(
+                        "The BaseEntityRef `%s` is not loaded, but is requested by the mapping `%s`.",
+                        entityRef.getUniqueObjectName(),
+                        mapping));
+            });
+        }
+        return property.getValue(entity);
+    }
+
     /**
```
andyHa, line 1084: Keep this as private method in SmartQuery, kill resolveJoins as it is never used - so probably we can just walk through mapping.getParent and repeatedly call getProperty().getValue() on the proper descriptors - so this all should be quite a lot simpler... On the actual call site, we know for sure, that the value is fetched, so no special checks are needed either...

src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -384,6 +387,28 @@ private List<E> queryNextBlock() {
...
+            Property property = entityDescriptor.getProperty(mapping.getName());
+            if (property instanceof BaseEntityRefProperty<?, ?, ?> ref) {
+                BaseEntityRef<?, ?> entityRef = ref.getEntityRef(entity);
+                return entityRef.getValueIfPresent().orElseThrow(() -> {
+                    return new IllegalArgumentException(Strings.apply(
+                            "The BaseEntityRef `%s` is not loaded, but is requested by the mapping `%s`.",
+                            entityRef.getUniqueObjectName(),
+                            mapping));
+                });
+            }
```
andyHa, line 409: isn't this only effective within the parent lookup(s) - can't we move this into an extra method? this whole method seems to do two things at once..

andyHa (changes requested):

src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -388,26 +387,29 @@ private List<E> queryNextBlock() {
...
 
-        private static Object getPropertyValue(EntityDescriptor entityDescriptor,
-                                               Mapping mapping,
-                                               Object entity) {
+        private static Object getPropertyValue(Mapping mapping, BaseEntity<?> entity) {
+            BaseEntity<?> parent = findParent(mapping, entity);
+            return parent.getDescriptor().getProperty(mapping.getName()).getValue(parent);
+        }
+
+        private static BaseEntity<?> findParent(Mapping mapping, BaseEntity<?> entity) {
```
andyHa, line 395:
```suggestion
        private BaseEntity<?> findParent(Mapping mapping, BaseEntity<?> entity) {
```

andyHa (commented):
no need for a static method...

src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -384,6 +386,31 @@ private List<E> queryNextBlock() {
 
             return effectiveQuery.where(OMA.FILTERS.gt(leftHandSide, rightHandSide)).queryList();
         }
+
+        private static Object getPropertyValue(Mapping mapping, BaseEntity<?> entity) {
```
qw3ry, line 390:
```suggestion
        private Object getPropertyValue(Mapping mapping, BaseEntity<?> entity) {
```

sabieber (approved):

idlira (approved):

PR #541, opened 2023-02-02
ES: Throws exception if truncate (deleteByQuery) encounters failures

cschierle:
Currently, we only forward version conflicts as `OptimisticLockException`. Now we also react to other failures (see [ES Docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query-api-response-body:~:text=to%20requests_per_second.-,failures,-Array%20of%20failures)) that might occur during a truncate operation by throwing a specific exception. The automatic retry is applied similar to the `OptimisticLockException` handling.

Fixes: [OX-9429](https://scireum.myjetbrains.com/youtrack/issue/OX-9429)


sabieber (approved):

andyHa (changes requested):

andyHa:
Shouldn't we rather check "version_conflicts" or "timed_out" and then throw an OL error. For everything else we can log and throw a handledexception just like all other APIs (or whatever we throw elsewhere)..

cschierle:
If `version_conflicts` is filled, those conflicts will also be listed in `failures`. Since we already catch OLEs via the HTTP response, I thought we only check, if there are any other failures. We could also check for `timed_out` though.

If we don't need the automatic retry, we could also switch to e.g. throwing a HandledException instead 👍

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -1465,18 +1465,31 @@ public void truncate() {
      * <p>
      * Use this for larger result sets where integrity and constraints do not matter or are managed manually.
      *
-     * @throws OptimisticLockException if one of the documents was modified during the runtime of the truncate
+     * @throws OptimisticLockException               if one of the documents was modified during the runtime of the truncate
+     * @throws sirius.kernel.health.HandledException if the {@linkplain LowLevelClient#deleteByQuery(String, String, JSONObject) deleteByQuery} request aborted due to any unrecoverable errors during the process
```
andyHa, line 1469: add line break (?)

fhaScireum (approved):

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -1465,18 +1465,31 @@ public void truncate() {
...
+            || deleteByQueryResponse.getIntValue("version_conflicts") > 0) {
+            throw new OptimisticLockException("Truncate timed out or had version conflicts:\n"
+                                              + deleteByQueryResponse.toJSONString());
+        }
+        if (deleteByQueryResponse.getJSONArray("failures") != null && !deleteByQueryResponse.getJSONArray("failures")
+                                                                                            .isEmpty()) {
+            throw Exceptions.createHandled()
+                            .withSystemErrorMessage("Truncate aborted due to unrecoverable error(s):\n"
+                                                    + deleteByQueryResponse.toJSONString())
+                            .handle();
```
andyHa, line 1491: KEY_XXX constants?

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -1465,18 +1465,31 @@ public void truncate() {
...
+        if (Boolean.TRUE.equals(deleteByQueryResponse.getBoolean("timed_out"))
+            || deleteByQueryResponse.getIntValue("version_conflicts") > 0) {
+            throw new OptimisticLockException("Truncate timed out or had version conflicts:\n"
+                                              + deleteByQueryResponse.toJSONString());
+        }
+        if (deleteByQueryResponse.getJSONArray("failures") != null && !deleteByQueryResponse.getJSONArray("failures")
+                                                                                            .isEmpty()) {
+            throw Exceptions.createHandled()
+                            .withSystemErrorMessage("Truncate aborted due to unrecoverable error(s):\n"
+                                                    + deleteByQueryResponse.toJSONString())
```
andyHa, line 1490: no we don't want to reveal this to the user.

Log to syslog and create a handled exception with a short error like "Truncate aborted due to unrecoverable error(s)."

idlira (approved):

sabieber (approved):

fhaScireum (approved):

idlira (approved):

src/main/java/sirius/db/es/ElasticQuery.java
```diff
@@ -1465,18 +1468,33 @@ public void truncate() {
...
+            throw new OptimisticLockException();
+        }
+        if (deleteByQueryResponse.getJSONArray(KEY_FAILURES) != null
+            && !deleteByQueryResponse.getJSONArray(KEY_FAILURES).isEmpty()) {
+            Elastic.LOG.SEVERE("Truncate aborted due to unrecoverable error(s):\n"
+                               + deleteByQueryResponse.toJSONString());
+            throw Exceptions.createHandled()
+                            .withSystemErrorMessage("Truncate aborted due to unrecoverable error(s)!")
+                            .handle();
+        }
```
andyHa, line 1497: shouldn't we switch this (move above OL warning)...

mkeckmkeck (approved):

src/main/java/sirius/db/es/ElasticQuery.java
cschierle: I think not.

> If `version_conflicts` is filled, those conflicts will also be listed in `failures`.

Version conflicts can be resolved by retrying the deletion for the conflicting documents in a new request, which is done by the automatic retry when throwing and catching the OLE. If we move the check for `failures` before the OLE and `failures` contains only version conflicts, we lose the option to retry. See https://discuss.elastic.co/t/version-conflict-while-using-delete-by-query/104831 for a sample response containing `version_conflicts` in `failures`.

PR #542, opened 2023-02-02
Relaxes "is active" aborts

mkeckmkeck:
Use case: 
- when a job gets aborted on UI, we want to quickly abort the processing
- but: the previous state prevented some quickly runnings db request to occur that might be required after the abort
- now we detect longer running requests, and only abort those
- means e.g. a queryOne or queryList can still run on a aborted job/task

### DIFF
--- a/src/main/java/sirius/db/jdbc/BaseSQLQuery.java
+++ b/src/main/java/sirius/db/jdbc/BaseSQLQuery.java
@@ -125,7 +125,7 @@ public abstract class BaseSQLQuery {
      * Executes the given query returning the first matching row.
      * <p>
      * If the resulting row contains a {@link Blob} an {@link OutputStream} as to be passed in as parameter
-     * with the name name as the column. The contents of the blob will then be written into the given
+     * with the name as the column. The contents of the blob will then be written into the given
      * output stream (without closing it).
      *
      * @return the first matching row for the given query or <tt>null</tt> if no matching row was found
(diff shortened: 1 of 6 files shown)


sabieber (approved):
Optional: rename `tc` and `rs` variables to a more descriptive variant

src/main/java/sirius/db/jdbc/BaseSQLQuery.java
```diff
@@ -95,8 +95,9 @@ public void iterateAll(Consumer<Row> consumer, @Nullable Limit limit) throws SQL
     protected void processResultSet(Predicate<Row> handler,
                                     Limit effectiveLimit,
                                     ResultSet resultSet,
-                                    TaskContext taskContext) throws SQLException {
-        while (resultSet.next() && taskContext.isActive()) {
+                                    TaskContext taskContext,
+                                    boolean longRunning) throws SQLException {
+        while (resultSet.next() && (taskContext.isActive() && longRunning)) {
```
andyHa, line 100: shoudln't we factor this out here entirely (we unvoke a handler returning a boolean anyway) - this way, we could only enforce this in interate, iterateAll and blockwise* without having the extra complexity in here. We might even handle this on a database independent level. Also "longRunning" isn't an ideal name, as otherwise "long running queries" are such, where the execution itself might be long running (even before results are seen) and thus should not belogged. This is however no indication on the result set size.

src/main/java/sirius/db/mongo/Finder.java
```diff
@@ -318,7 +318,7 @@ private void processCursor(MongoIterable<Document> cursor, Predicate<Doc> proces
             }
 
             boolean keepGoing = processor.test(new Doc(doc));
-            if (!keepGoing || !taskContext.isActive()) {
+            if (!keepGoing || (!taskContext.isActive() && longRunning)) {
```
andyHa, line 321: This is most probably the "wrong" kind of long running...

andyHa (changes requested):

PR #549, opened 2023-02-28
SIRI-732: enables a blockwise query with null values

bkositza:
In MySQL/MariaDB, NULL is considered as a 'missing, unkonwn value'. Any arithmetic comparison with NULL returns false e.g. NULL != 'any' returns false. Therefor we need to return Boolean#FALSE instead of NULL because false != 'any' will be true.

### DIFF
--- a/src/main/java/sirius/db/jdbc/constraints/CompoundValue.java
+++ b/src/main/java/sirius/db/jdbc/constraints/CompoundValue.java
@@ -76,13 +76,30 @@ public class CompoundValue {
                 sqlRepresentation.add(compiler.translateColumnName(mapping));
             } else {
                 sqlRepresentation.add("?");
-                parameters.add(component);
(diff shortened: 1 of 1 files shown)


qw3ry:
I don't know about this fix. It only applies to CompoundValues which are fairly new and probably not widely used, but it also changes the behavior of the database intransparently...

Can we instead fix this in `SmartQuerySpliterator#getPropertyValue`?

andyHa:
> I don't know about this fix. It only applies to CompoundValues which are fairly new and probably not widely used, but it also changes the behavior of the database intransparently...
> 
> Can we instead fix this in `SmartQuerySpliterator#getPropertyValue`?

Once we reach a null column, shouldn't we just filter on  column == null (or actually column is null) as there is nothing left to order by?

PR #553, opened 2023-03-17
Enables a blockwise query with null values

bkositza:
In MySQL/MariaDB, NULL is considered as a 'missing, unkonwn value'. Any arithmetic comparison with NULL returns false e.g. NULL != 'any' returns false.

Therefore, comparisons with NULL values must be treated specially when it comes to sorting in a blockwise query.

The problem was that we try to filter on everything larger than the last value of the last block when go on to the next block. But a greater-than filter does not work if the last value of the last block is NULL (in any sorting column). This resulted in the next block not containing any values, so we never reached the next block and always loaded too few records.

This PR tries to fix that. Please check the logic behind the changes very carefully. 
This PR definitely needs to be well tested later on as well.

Fixes: SIRI-732


src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -366,25 +366,54 @@ private List<E> queryNextBlock() {
...
+                previousSortingColumns.put(sortColumn, value);
             }
+            return effectiveQuery.where(sortingFilterConstraint).queryList();
+        }
 
-            return effectiveQuery.where(OMA.FILTERS.gt(leftHandSide, rightHandSide)).queryList();
+        /**
+         * Creates a sql constraint for sorting purposes.
+         * </p>
+         * In MySQL/MariaDB, NULL is considered as a 'missing, unkonwn value'. Any arithmetic comparison with NULL
```
andyHa, line 388: is this mysql specific or according to SQL standard? if this would be specific we should tie this to a database capability...

src/main/java/sirius/db/jdbc/SmartQuery.java
```diff
@@ -366,25 +366,54 @@ private List<E> queryNextBlock() {
...
+         * returns false e.g. NULL != 'any' returns false.
+         * Therefore, comparisons with NULL values must be treated specially.
+         *
+         * @param sortAscending decides whether the sorting direction is descending or ascending
+         * @param column the column to be used for sorting
+         * @param value the value of the column
+         * @param previousSortingColumns all columns that should be sorted before the current one
+         * @return {@link SQLConstraint} which can be used to map a level of sorting.
+         */
+        private SQLConstraint createSqlConstraintForSortingColumn(boolean sortAscending, Mapping column, Object value,
```
andyHa, line 398: can we provide a unit test / integration test (whatever works) for this whole thing?

src/main/java/sirius/db/jdbc/SmartQuery.java
qw3ry: This is basically SQL Standard, but there are also implementation specific differences...

qw3ry: More precisely: `NULL != 'any'` => false is SQL standard. However, the sort order with null values (nulls first vs last) is implementation specific

andyHa (approved):

qw3ry (approved):

sabieber (approved):

idlira (approved):

mkeckmkeck (approved):

PR #558, opened 2023-04-03
Properly encapsulates qdrant matches

andyHa:
We commonly need to extract multiple payloads and also want to have access to the score of each match.

### DIFF
--- a/src/main/java/sirius/db/qdrant/Match.java
+++ b/src/main/java/sirius/db/qdrant/Match.java
@@ -21,16 +21,14 @@ public class Match {
     private final float score;
 
     /**
-     * Creates a new match with the given id, score and payload.
+     * Creates a new match from the given JSON response
      *
-     * @param id      the id of the matched point
-     * @param score   the score or distance of the match
-     * @param payload the payload of the matched point
+     * @param match the match as returned by qdrant
      */
-    public Match(String id, float score, JSONObject payload) {
-        this.id = id;
-        this.score = score;
-        this.payload = payload;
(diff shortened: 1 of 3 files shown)


mko-sci (approved):

src/main/java/sirius/db/qdrant/Match.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class Match {
+
+    private final String id;
+    private final JSONObject payload;
+
+    private final float score;
+
+    /**
+     * Creates a new match with the given id, score and payload.
```
sabieber, line 24: **Optional:** `ID` should be uppercase in english

src/main/java/sirius/db/qdrant/Search.java
```diff
@@ -90,16 +92,17 @@ public <T> List<T> queryPayload(Class<T> fieldType, String field, int maxResults
...
-
-        List<T> payloadData = new ArrayList<>(points.size());
-        for (JSONObject point : points.toJavaList(JSONObject.class)) {
-            payloadData.add(fieldType.cast(point.getJSONObject("payload").get(field)));
-        }
-
-        return payloadData;
+        return points.stream()
+                     .map(JSONObject.class::cast)
+                     .map(point -> new Match(point.getString("id"),
```
sabieber, line 102: **Optional:** A constructor that takes a `JSONObject` could be added to internalize parsing/fetching the correct fields

sabieber (approved):

PR #574, opened 2023-06-14
Permits EntityRefs to fetch values using a secondary node

idlira:
...of a clustered DB setup.

Note that calling EntityRef.fetchFromSecondary always performs a lookup and the value returned (if any) will not be cached.

### DIFF
--- a/src/main/java/sirius/db/mixing/types/BaseEntityRef.java
+++ b/src/main/java/sirius/db/mixing/types/BaseEntityRef.java
@@ -224,6 +224,8 @@ public abstract class BaseEntityRef<I extends Serializable, E extends BaseEntity
 
     /**
      * Performs the lookup of the entity with the given id using a secondary node of a DB cluster.
+     * <p>
+     * DBs not supporting this falls back to {@link #find(Class, Serializable)}.
      *
      * @param type the type to search for
      * @param id   the id to lookup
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/db/mixing/types/BaseEntityRef.java
```diff
@@ -203,6 +203,15 @@ public Optional<E> getValueIfPresent() {
      */
     protected abstract Optional<E> find(Class<E> type, I id);
 
+    /**
+     * Performs the lookup of the entity with the given id using a secondary node of a DB cluster.
```
sabieber, line 210: **Optional:** Should mention that this does the same as `find` on db systems that don't support fetching from secondary nodes.

src/main/java/sirius/db/mixing/types/BaseEntityRef.java
```diff
@@ -302,4 +305,16 @@ public String toString() {
             return "(empty)";
         }
     }
+
+    private E fetchValue(boolean useSecondary) {
+        if (value == null && id != null) {
+            Optional<E> entity = useSecondary ? findInSecondary(type, id) : find(type, id);
+            if (entity.isPresent()) {
+                value = entity.get();
```
sabieber, line 313: In my opinion we should not store/cache the value when it was fetched from secondary but just return it. This could pollute a cache with outdated data.

src/main/java/sirius/db/mixing/types/BaseEntityRef.java
```diff
@@ -180,6 +180,25 @@ public E fetchValue() {
...
+     * @return the entity being referenced or <tt>null</tt> if no entity is referenced.
+     */
+    public E fetchValueFromSecondary() {
+        if (value == null && id != null) {
+            Optional<E> entity = findInSecondary(type, id);
+            if (entity.isPresent()) {
+                return entity.get();
+            } else {
+                id = null;
+            }
```
sabieber, line 197: Would also not clear the ID in this case. It could be that a newly inserted entity just isn't propagated through the cluster yet.

mkeckmkeck (approved):

PR #576, opened 2023-06-23
Project update and code cleanup

andyHa:


src/main/java/sirius/db/mixing/query/constraints/FilterFactory.java
```diff
@@ -376,7 +376,7 @@ public C not(@Nullable C constraint) {
      * @param constraint the constraint to invert
      * @return the inverted constraint
      */
-    @Nonnull
+    @Nullable
```
sabieber, line 379: **Question:** Are all these nullability changes correct? How can a constraint inversion of a nonnull constraint return a nullable constraint?

src/main/java/sirius/db/text/PatternExtractProcessor.java
```diff
@@ -19,7 +21,9 @@
  */
 public class PatternExtractProcessor extends ChainableTokenProcessor {
 
-    private static final Pattern EXTRACT_EMAILS = Pattern.compile("(\\p{Alnum}.+)@(.+)$");
+    @SuppressWarnings("java:S5852")
+    @Explain("This regex cannot cause catastrophic backtracking.")
```
jakobvogel, line 25: Is it possible to fix this with possessive quantifiers, rather than suppressing the warning?

jakobvogel (approved):
Subject to @sabieber's question 😉

src/main/java/sirius/db/mixing/query/constraints/FilterFactory.java
andyHa: e.g. an And constraint with zero clauses might be optimized down to "null" and then inverted to "null" - we might also mark the parameter as nullable aswell

sabieber: But isn't the parameter also nullable then?

src/main/java/sirius/db/text/PatternExtractProcessor.java
andyHa: We just had a production issue, which might correlate to this regex - maybe we revise again...

PR #582, opened 2023-07-18
Allows to define shared validators for properties

sabieber:
This allows to define common validations that are used on fields across multiple entities. Example could be something like email validation.
The alternative of wrapping single fields in composites can be kind of cumbersome and lead to ugly field names and fragmentation.

With the new `ValidatedBy` annotation a validator can be named that should be called in the validation and before-save phases of the property/entity lifecycle. The validator has to be globally registered and implement the `PropertyValidator` interface while specifying a unique name.


idlira (approved):

mko-sci (approved):

andyHa (approved):

sabieber:
> Maybe we could provide a standard validator like "phone number" or "email address" - this should be present already (either in -db or -biz)

Yes, my plan is to provide some in biz after this is accepted :)

PR #591, opened 2023-09-07
Introduces a non-strict validation mode

mkeckmkeck:
Some fields do not have a strong need need for validity. E.g. Legacy data for phone numbers like "123 or 456" might be acceptable. 
This is where non-strict validation comes into place. Non-strict validation only applies the validation on changed field values. Actually untouched legacy data (that might be invalid) will not be validated and so does not prevent the update of other fields. 

Fixes: SIRI-869

### DIFF
--- a/src/main/java/sirius/db/mixing/Property.java
+++ b/src/main/java/sirius/db/mixing/Property.java
@@ -136,6 +136,12 @@ public abstract class Property extends Composable {
     protected boolean nullable;
 
     protected PropertyValidator propertyValidator;
+
+    /**
+     * Determines if the {@link #propertyValidator} should be also executed when the value has not changed.
+     * <p>
+     * This is only relevant when a validator was specified via {@link ValidatedBy}.
+     */
     protected boolean strictValidation = true;
 
     /**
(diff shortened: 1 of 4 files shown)


fhaScireum (approved):

mko-sci (approved):

src/main/java/sirius/db/mixing/Property.java
```diff
@@ -716,12 +724,14 @@ protected final void onBeforeSave(Object entity) {
...
+        if (propertyValidator != null && (strictValidation || valueChanged)) {
+            // Only enforce validity if the value actually changed or in strict validation mode
             propertyValidator.beforeSave(this, getValue(entity));
         }
 
-        if (entity instanceof BaseEntity<?> && (((BaseEntity<?>) entity).isNew() || ((BaseEntity<?>) entity).isChanged(
-                nameAsMapping))) {
+        if (valueChanged) {
             // Only enforce uniqueness if the value actually changed...
             checkUniqueness(entity, propertyValue);
```
idlira, line 736: Maybe I'm blind, but uniqueness check should be called independent on validators (strict or not).

idlira (changes requested):

src/main/java/sirius/db/mixing/Property.java
mkeckmkeck: It is called independent. There are two ifs, sharing the extracted valueChanged variable.

idlira (approved):

PR #603, opened 2023-12-14
update some dependencies.

ymo-sci:
httpclient is now included in the newer elasticsearch-rest-client in the until now provided version so we can remove it. setMaxWaitMillis is deprecated so we use setMaxWait now (which is internally the same)

- fixes: SIRI-920

### DIFF
--- a/src/main/java/sirius/db/jdbc/Database.java
+++ b/src/main/java/sirius/db/jdbc/Database.java
@@ -353,7 +353,7 @@ public class Database {
             ds.setMaxIdle(maxIdle);
             ds.setTestOnBorrow(testOnBorrow);
             ds.setValidationQuery(validationQuery);
-            ds.setMaxWait(Duration.ofMillis(1000));
+            ds.setMaxWait(Duration.ofSeconds(1));
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/db/jdbc/Database.java
```diff
@@ -353,7 +353,7 @@ private void initialize() {
             ds.setMaxIdle(maxIdle);
             ds.setTestOnBorrow(testOnBorrow);
             ds.setValidationQuery(validationQuery);
-            ds.setMaxWaitMillis(1000);
+            ds.setMaxWait(Duration.ofMillis(1000));
```
sabieber, line 356: **Optional:** Why not `Duration.ofSeconds(1)`?

src/test/kotlin/sirius/db/testutil/MongoMocks.kt
```diff
@@ -8,7 +8,7 @@
 
 package sirius.db.testutil
 
-import io.mockk.spyk
+import org.mockito.Mockito.spy
```
sabieber, line 11: **Question:** But we still want to be able to use MockK?? Why is this change necessary?

mkeckmkeck (approved):

sabieber (approved):

jmuscireum (approved):

PR #648, opened 2024-04-09
Optimizes some system generates queries based on casing annotations

mkeckmkeck:
Uses lowercase and uppercase field annotations to optimize queries

- When all values stored in a field are stored e.g. lowercased, we do not need
  to query case-insensitively. This can give a small speedup.
- When all values in a field are stored e.g. lowercased, we do not need
  to query with an value containing upper case characters. We apply auto-magical
  case adaption on some cases where the framework generates queries. So we create
  the chance of actually finding an result, even if the input did not match the actual db field.
  This is relevant, e.g. for the Sirius biz virtual filesystem backend ui. There, the normalized
  filename field of a sql blob containing only lowercase characters might get queried by user input
  containing the exact filename. By this adaption, we are able to find the blob.
- For regular API usage, such magic from framework is not desired. If required, apply it by
  yourself when creating queries.

### DIFF
--- a/src/main/java/sirius/db/mixing/query/QueryCompiler.java
+++ b/src/main/java/sirius/db/mixing/query/QueryCompiler.java
@@ -11,6 +11,8 @@ package sirius.db.mixing.query;
 import sirius.db.mixing.EntityDescriptor;
 import sirius.db.mixing.Mapping;
 import sirius.db.mixing.Property;
+import sirius.db.mixing.annotations.LowerCase;
+import sirius.db.mixing.annotations.UpperCase;
 import sirius.db.mixing.properties.BaseEntityRefListProperty;
 import sirius.db.mixing.properties.BaseEntityRefProperty;
 import sirius.db.mixing.properties.LocalDateProperty;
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -789,4 +792,22 @@ private C parseTag() {
...
+     */
+    protected Optional<String> getCaseOptimizedValue(Mapping field, String value) {
+        Property property = resolveProperty(field.toString()).getSecond();
+        if (property.isAnnotationPresent(LowerCase.class)) {
+            return Optional.of(value.toLowerCase());
+        }
+        if (property.isAnnotationPresent(UpperCase.class)) {
+            return Optional.of(value.toUpperCase());
+        }
+        return Optional.empty();
```
sabieber, line 810: **Optional:** Why not use a plain String as return type and give back the original value in this case?

sabieber (approved):

src/main/java/sirius/db/mixing/query/QueryCompiler.java
mkeckmkeck: Using the optional I can easily decide if an optimization had occured. Then I can react on it accordingly, e.g. if value was lowercased, a case-insensitive query is not required. I made some quick and not reliable measuments using sql console with lower(value) vs value, and this can bring up to 10% performance gain.

idlira (approved):

mko-sci (approved):

jakobvogel (approved):

PR #649, opened 2024-04-16
Reference to Raw Mongo Object

jakobvogel:
### Description

When parsing a Mongo object into a Mango entity, we now retain a reference to the raw document. This is similar to our implementation of the Elastic adapter.


sabieber (approved):

andyHa (approved):

jakobvogel:
The second commit was necessary following late feedback in `scireum-dev`. The same commit makes the entire solution unattractive for the intended use in MIO.

PR #668, opened 2024-12-13
Improves the BaseEntityCache and BaseEntityRef

idlira:
### Description

- BaseEntityCache
Always use the ID stored in the reference to retrieve an item from the cache. Previously, if the reference was pre-loaded, that value was delivered instead.
- BaseEntityRef
`fetchValue()` is now deprecated and replaced by `fetchCachedValue()` + `forceFetchValue()`. Same for the `*FromSecondary` variants.

Marked as **BREAKING** since usages of fetchValue() and fetchValueFromSecondary() in tagliatelle templates might lead to build warnings.

To keep the previous semantics, simply replace `fetchValue()` by `fetchCachedValue()` and `fetchValueFromSecondary()` by `fetchCachedValueFromSecondary()`

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1039](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1039)

### Checklist

### DIFF
--- a/src/main/java/sirius/db/mixing/types/BaseEntityRef.java
+++ b/src/main/java/sirius/db/mixing/types/BaseEntityRef.java
@@ -200,12 +200,25 @@ public abstract class BaseEntityRef<I extends Serializable, E extends BaseEntity
      *
      * @return the entity being referenced or <tt>null</tt> if no entity is referenced.
      */
-    public E fetchValueFromSecondary() {
+    public E fetchCachedValueFromSecondary() {
         if (value == null && id != null) {
             Optional<E> entity = findInSecondary(type, id);
             return entity.orElse(null);
         }
         return value;
+    }
+
+    /**
+     * Returns the effective entity object which is referenced using a secondary node of a DB cluster.
+     * <p>
+     * Note, this values returned by this method will never be cached and a new lookup is always performed.
+     *
+     * @return the entity being referenced or <tt>null</tt> if no entity is referenced.
+     * @deprecated use {@link #fetchCachedValueFromSecondary()} instead
+     */
+    @Deprecated
+    public E fetchValueFromSecondary() {
+        return fetchCachedValueFromSecondary();
     }
 
     /**
--- a/src/main/java/sirius/db/mixing/types/BaseEntityRef.java
+++ b/src/main/java/sirius/db/mixing/types/BaseEntityRef.java
@@ -167,7 +167,7 @@ public abstract class BaseEntityRef<I extends Serializable, E extends BaseEntity
      * Note, this might cause a database lookup if the entity is not prefetched.
      *
      * @return the entity being referenced or <tt>null</tt> if no entity is referenced.
-     * @deprecated use {@link #fetchCachedValue()} instead
+     * @deprecated use {@link #fetchCachedValue()} or {@link #forceFetchValue()} instead
      */
     @Deprecated
     public E fetchValue() {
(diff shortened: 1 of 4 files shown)


src/test/kotlin/sirius/db/jdbc/SmartQueryKotlinTest.kt
```diff
@@ -31,24 +31,24 @@ class SmartQueryKotlinTest {
     @Test
     fun `queryList returns all entities`() {
         val smartQueryTestEntity = oma.select(SmartQueryTestEntity::class.java)
-                .orderAsc(SmartQueryTestEntity.TEST_NUMBER)
+            .orderAsc(SmartQueryTestEntity.TEST_NUMBER)
```
jakobvogel, line 34: Looks as if we need to optimise our Kotlin formatting rules 😇

src/test/kotlin/sirius/db/jdbc/SmartQueryKotlinTest.kt
```diff
@@ -283,7 +283,10 @@ class SmartQueryKotlinTest {
         )
         val result = smartQueryTestChildEntity.queryList()
 
-        assertEquals(listOf("Parent 2"), result.stream().map { x -> x.name }.collect(Collectors.toList()))
+        assertEquals(
+            listOf("Parent 2", "Parent 3"),
+            result.stream().map { x -> x.name.substring(0, 8) }.collect(Collectors.toList())
```
jakobvogel, line 288: - Add a comment explaining this?
- Protect against `x.name.length` < 8?
- Compare sets (or sorted lists) if we cannot guarantee the order?

jakobvogel (approved):

src/test/kotlin/sirius/db/jdbc/SmartQueryKotlinTest.kt
idlira: I've added a comment about why we substring.
For the other comments: The data used in these test cases are inserted during the class setup, therefore controlled data, which means:
1. `x.name.length < 8` cannot happen
2. The order of items will be guaranteed by the ID (they are inserted in order). What we cannot guarantee is the oder in which tests are executed.

src/main/java/sirius/db/mixing/types/BaseEntityRef.java
```diff
@@ -180,21 +193,55 @@ public E fetchValue() {
...
     }
 
+    /**
+     * Returns the effective entity object which is referenced by forcing a fresh database lookup.
+     *
+     * @return the entity being referenced or <tt>null</tt> if no entity is referenced.
+     */
+    public E forceFetchValue() {
+        value = null;
+        return fetchCachedValue();
```
andyHa, line 203: Optional: Would feel a bit natural to bury the find lookup here and to let fetchCachedValue fallback to forceFetchValue insted of the other way around...

src/main/java/sirius/db/mixing/types/BaseEntityRef.java
```diff
@@ -180,21 +193,55 @@ public E fetchValue() {
...
 
+    /**
+     * Returns the effective entity object which is referenced by forcing a fresh database lookup using a secondary node
+     * of a DB cluster.
+     *
+     * @return the entity being referenced or <tt>null</tt> if no entity is referenced.
+     */
+    public E forceFetchValueFromSecondary() {
+        value = null;
+        return fetchCachedValueFromSecondary();
```
andyHa, line 229: s.a.

src/test/kotlin/sirius/db/jdbc/SmartQueryKotlinTest.kt
sabieber: They are (or rather should be when the IDE is configured correctly) defined by the Kotlin language themselves. It is highly discouraged to stray away from it and create a customized formatting ruleset.
But this formatting looks like the language default isn't applied here.
Maybe we can change something in sirius-parent to make usage easier.

sabieber (approved):

scireum-mbo (approved):

mkeckmkeck (approved):

PR #673, opened 2025-02-17
introduce case-sensitive parameter for SQL QueryFields.

ymo-sci:
### Description

also switch mariaDB collation in tests to case-sensitive as we mostly use that to test it properly

earlier we used _ci collations and this wasn't a problem then...

The use case ist e.g. searching a username by startsWith. until now a user named "Max Mustermann" cannot be found by typing "max" oder "mustermann", but via "max*" e.g. so we introduce a new method for searching SQL in LIKE or PREFIX mode case-insensitive, but that won´t use an existing index then so only should be called on limited values.

on Mongo + Elastic contraints the logic is already case-insensitve.

Technically BREAKING for Implementations of sirius.db.mixing.query.QueryCompiler

### Additional Notes

- This PR fixes or works on following ticket(s): [SE-14367](https://scireum.myjetbrains.com/youtrack/issue/SE-14367)
- This PR is related to PR: 

### Checklist

- [ ] Patch Tasks: Is local execution of Patch Tasks necessary? If so, please also mark the PR with the tag.

### DIFF
--- a/src/main/java/sirius/db/mixing/query/QueryCompiler.java
+++ b/src/main/java/sirius/db/mixing/query/QueryCompiler.java
@@ -502,6 +502,7 @@ public abstract class QueryCompiler<C extends Constraint> {
             for (QueryField field : searchFields) {
                 fieldConstraints.add(compileSearchToken(field.getField(),
                                                         QueryField.Mode.EQUAL,
+                                                        field.isCaseSensitive(),
                                                         token.getValue().toString()));
             }
             constraints.add(factory.or(fieldConstraints));
@@ -509,7 +510,10 @@ public abstract class QueryCompiler<C extends Constraint> {
             for (String word : token.getValue().toString().split("\\s")) {
                 List<C> fieldConstraints = new ArrayList<>();
                 for (QueryField field : searchFields) {
-                    fieldConstraints.add(compileSearchToken(field.getField(), field.getMode(), word));
+                    fieldConstraints.add(compileSearchToken(field.getField(),
+                                                            field.getMode(),
+                                                            field.isCaseSensitive(),
+                                                            word));
                 }
                 constraints.add(factory.or(fieldConstraints));
             }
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/db/mixing/query/QueryField.java
```diff
@@ -55,20 +57,50 @@ public static QueryField eq(Mapping field) {
...
+        return new QueryField(field, Mode.PREFIX, true);
+    }
+
+    /**
+     * Informs the compiler, that by default an expanding search like a case-insensitive prefix query may be generated for this field.
+     * <p>
+     * Depending on the underlying database an index might be used to support this query. If the database cannot
+     * execute prefix queries it may fall back to an equals-constraint.
+     * <p>
+     * Caveat: This method is case-insensitive and therefore will <b>not use an index</b> and should only be used for reasonably sized datasets.
```
fhaScireum, line 97: The two sentences clash with each other concerning whether an index **might be used** or whether an index will **not be used at all**.

src/main/java/sirius/db/mixing/query/QueryCompiler.java
```diff
@@ -520,12 +524,13 @@ protected C compileDefaultSearch(List<QueryField> searchFields, FieldValue token
...
-     * @param mode  the mode to use
-     * @param value the value to search for
+     * @param field         the field to search in
+     * @param mode          the mode to use
+     * @param caseSensitive determines if the search should be case-sensitive
+     * @param value         the value to search for
      * @return an appropriate constraint for the given parameters
      */
-    protected abstract C compileSearchToken(Mapping field, QueryField.Mode mode, String value);
+    protected abstract C compileSearchToken(Mapping field, QueryField.Mode mode, boolean caseSensitive, String value);
```
idlira, line 533: This is changing the signature of an existing API, thus making this PR technically BREAKING.

mko-sci (approved):

ymo-sci:
Closed:
- we don't want to break the API
- the implementations are very different...
- Problem in this case can be solved via Query.contains instead of startsWith with only rare cases like "nick" will also find "yannick"
- The use of PrefixSearchableEntity is not possible because SQLQueryCompiler does not supports it right. If you put firstname and lastname to the Searchable fields, only "max" will find a result but not "mustermann" because the startswith in a stringlist in SQL is "max,mustermann"

PR #691, opened 2025-07-21
Updates the ClickHouse Driver to 0.9.0

idlira:
### Description

_... The ClickHouse JDBC driver is like a box of chocolates. You never know what breaking changes you are going to get... (Forrest Gump)_ 

**Note** The ClickHouse libraries will now be included as runtime libs via sirius-db.

Pity that the JDBC Implementation is still so crappy. For the timestamp issues they should respect the JDBC connection parameters like **useTimezone** and the like. 😟

- DB type `Date` has no timezone information and applications running in timezones higher than server's timezone will result in dates being stored in the previous day since it expects these in UTC: https://clickhouse.com/docs/sql-reference/functions/type-conversion-functions#fromunixtimestamp64nano

```
System Clock 2025-07-21 14:00:00 GMT+2
LocalDate: 2025-07-21
UTC of the date: 2025-07-20 22:00:00 UTC
Effectively stored: 2025-07-20
```

- DB type `DateTime` highest resolution is 1 second, so we cannot let the server parse a date time from a string and must delegate to the proper method, also considering the UTC issue as stated above. https://clickhouse.com/docs/sql-reference/data-types/datetime

- Resolved issue saving string lists
- Improved test cases

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1084](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1084)
- This PR replaces the PR: https://github.com/scireum/sirius-db/pull/690

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,6 @@
             <groupId>com.clickhouse</groupId>
             <artifactId>clickhouse-jdbc</artifactId>
             <version>0.9.0</version>
-            <scope>test</scope>
         </dependency>
     </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -89,8 +89,7 @@
         <dependency>
             <groupId>com.clickhouse</groupId>
             <artifactId>clickhouse-jdbc</artifactId>
-            <version>0.6.5</version>
-            <classifier>http</classifier>
+            <version>0.9.0</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/src/main/java/sirius/db/jdbc/Capability.java
+++ b/src/main/java/sirius/db/jdbc/Capability.java
@@ -77,7 +77,12 @@ public enum Capability {
      * Some, e.g. mariaDB or MySQL, lists null values before non-null values. Oracle and postgreSQL do it the other way
      * around.
      */
-    NULLS_FIRST;
+    NULLS_FIRST,
+
+    /**
+     * Signalizes that {@link sirius.db.mixing.properties.InstantProperty} values are stored without nanoseconds (1 second resolution).
+     */
+    INSTANT_WITHOUT_NANOS;
 
     /**
      * Contains the default capabilities of unknown databases.
@@ -108,5 +113,6 @@ public enum Capability {
     /**
      * Contains the capabilities of a Clickhouse database
      */
-    public static final Set<Capability> CLICKHOUSE_CAPABILITIES = Collections.unmodifiableSet(EnumSet.of(LIMIT, LISTS));
+    public static final Set<Capability> CLICKHOUSE_CAPABILITIES =
+            Collections.unmodifiableSet(EnumSet.of(LIMIT, LISTS, INSTANT_WITHOUT_NANOS));
 }
--- a/src/main/java/sirius/db/jdbc/Capability.java
+++ b/src/main/java/sirius/db/jdbc/Capability.java
@@ -80,18 +80,6 @@ public enum Capability {
     NULLS_FIRST,
 
     /**
-     * Used by {@link sirius.db.mixing.properties.LocalDateProperty} that the offset between the JVM's timezone and the
-     * server's timezone must be considered when storing a {@link java.time.LocalDate}.
-     * <p>
-     * The local date is obtained by stripping time from the current time in the current timezone. If the server's timezone is
-     * before the JVM's timezone, the date is shifted by one day.
-     * <p>
-     * For example, a {@link java.time.LocalDate} of 2023-10-01 00:00:00 in GMT+2 results in an actual date of 2023-09-30 22:00:00 in UTC,
-     * and this date will be stored in the database without its time information.
-     */
-    TIMEZONE_OFFSET,
-
-    /**
      * Signalizes that {@link sirius.db.mixing.properties.InstantProperty} values are stored without nanoseconds (1 second resolution).
      */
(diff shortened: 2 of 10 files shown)


src/main/java/sirius/db/jdbc/Capability.java
```diff
@@ -77,7 +77,12 @@ public enum Capability {
      * Some, e.g. mariaDB or MySQL, lists null values before non-null values. Oracle and postgreSQL do it the other way
      * around.
      */
-    NULLS_FIRST;
+    NULLS_FIRST,
+
+    /**
+     * Signalizes that {@link sirius.db.mixing.properties.InstantProperty} values are stored without nanoseconds (1 second resolution).
+     */
+    INSTANT_WITHOUT_NANOS;
```
sabieber, line 97: **Optional:** Maybe "invert" the capability? The class javadoc suggestions to me that it reflects something some dbs support and some don't, so maybe call it `INSTANT_NANO_RESOLUTION` or `PRECISE_INSTANTS`. Then it could be enabled by default and disabled for ClickHouse

src/main/java/sirius/db/mixing/properties/InstantProperty.java
```diff
@@ -83,11 +91,23 @@ protected Object transformToJDBC(Object object) {
...
 
     @Override
     public void contributeToTable(Table table) {
         table.getColumns().add(new TableColumn(this, Types.TIMESTAMP));
     }
+
+    private boolean hasCapabilityInstantWithoutNanos() {
+        if (dbHasCapabilityInstantWithoutNanos == null) {
+            dbHasCapabilityInstantWithoutNanos =
+                    oma.getDatabase(descriptor.getRealm()).hasCapability(Capability.INSTANT_WITHOUT_NANOS);
```
sabieber, line 109: **Question:** Is this an expensive operation we have to cache?

src/main/java/sirius/db/jdbc/Database.java
```diff
@@ -119,6 +123,20 @@ protected Database(String name) {
                                Formatter.create(profile.get(KEY_VALIDATION_QUERY).asString()).setDirect(ctx).format() :
                                ext.get(KEY_VALIDATION_QUERY).asString();
         this.testOnBorrow = Strings.isFilled(validationQuery);
+        this.timeZoneOffset = calculateTimeZoneOffset(ext.get(KEY_SERVER_TIME_ZONE).isFilled() ?
+                                                      ext.get(KEY_SERVER_TIME_ZONE).asString() :
+                                                      profile.get(KEY_SERVER_TIME_ZONE).asString());
```
sabieber, line 128: **Optional:** Maybe move the extraction into the helper function to increase readability slightly?

src/main/java/sirius/db/mixing/properties/LocalDateProperty.java
```diff
@@ -160,4 +178,16 @@ protected void determineDefaultValue() {
...
+
+    private int fetchTimezoneOffset() {
+        if (timeZoneOffset == null) {
+            Database database = oma.getDatabase(descriptor.getRealm());
+            if (database.hasCapability(Capability.TIMEZONE_OFFSET)) {
+                timeZoneOffset = database.getTimeZoneOffset();
+            } else {
+                timeZoneOffset = 0;
+            }
+        }
```
sabieber, line 190: **Optional:** Maybe move the capability check into the offset calculation logic if possible

src/main/java/sirius/db/jdbc/Capability.java
idlira: I don't know what all other flavors of support in terms of milli/nanos and what their drivers already solve from a given Instant (which is also rare, most of our entities uses LocalDateTime which stores a long).
So I prefer to actively activate for Clickhouse instead of defaulting on others.

src/main/java/sirius/db/mixing/properties/LocalDateProperty.java
idlira: I decided to not have it attached to the capability. An offset calculated during DB init is of no harm and its calculated as soon as the serverTimeZone is defined. The TIMEZONE_OFFSET capability will simply make use of it, but others could have access to that number if necessary.

mkeckmkeck (approved):
The TIMEZONE_OFFSET handling looks a little quirky. Would there be other options to handle this?

pom.xml
```diff
@@ -89,9 +89,7 @@
         <dependency>
             <groupId>com.clickhouse</groupId>
             <artifactId>clickhouse-jdbc</artifactId>
-            <version>0.6.5</version>
-            <classifier>http</classifier>
-            <scope>test</scope>
```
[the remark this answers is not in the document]
idlira: If I don't do it, I cannot use it in the `instance of` comparison.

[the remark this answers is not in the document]
idlira: But it is again in sirius-biz, so now we can remove from there since it will get exposed via DB (which in my opinion is somehow more logical)

mkeckmkeck (approved):

scireum-mbo (approved):

pom.xml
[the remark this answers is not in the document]
idlira: As decided, we will move the dependencies as runtime to sirius-db and remove from other modules.

oolscireum (approved):

fhaScireum (approved):

